Issue #16, opened 2023-11-09
Load historical raw data from PeMS clearinghouse


ian-r-rose:
The PeMS clearinghouse has ~20 years of historical data at 30-second frequency. We would like to load it into S3, and then Snowflake, including:

1. The raw count data
2. The station metadata
3. The station XML operational status (maybe)

ian-r-rose:
Closing as complete, follow-up work in setting up snowpipe is tracked by #19

PR #18, opened 2023-11-16
PeMS Clearinghouse historical load

ian-r-rose:
Work in progress for fixing #16


ian-r-rose:
Okay, this is ready!

AeriShan-ODI (approved):
Not that I've tested it or anything crazy like that but it all looks very sensible

PR #20, opened 2023-11-27
Set up external stage and pipes

### LINKED ISSUE #19: Create stage and pipe infrastructure for loading historical PeMS data to Snowflake
In #18 we are loading historical data to S3. The next step would be to establish the bucket as an external stage and load it to Snowflake, including pipes for updates. Prior work on this is in https://github.com/cagov/data-infrastructure/issues/69

ian-r-rose:
Fixes #19

This sets up the relevant infrastructure in both AWS and Snowflake for auto-loading data using the project raw S3 bucket as an external stage. I think this really starts to show some of the benefits of using Terraform for this! We're creating infrastructure in both, and using configuration values from one in the other.

This implements the approach given in [these docs](https://docs.snowflake.com/user-guide/data-load-snowpipe-auto-s3#step-5-grant-the-iam-user-permissions-to-access-bucket-objects), and follows the proof-of-concept I wrote some months ago in https://github.com/cagov/data-infrastructure/pull/108.


melanie-logan (approved):
Reviewed and approved.

PR #25, opened 2023-12-29
Missing Data: An Initial Exploration

mbobra:
**Purpose**: Identify missing data in the Caltrans PeMS dataset. The nature of the missing data will inform any future imputation solutions.

**Changed Files**

Added additional packages to the environment
- [x] `pyproject.toml` Added python-dotenv
- [x] `poetry.lock` Added the associated lock file

Added a Jupyter notebook for initial exploration, `pems/missing_data.ipynb`. 
  - [x] Set up the environment and connection to Snowflake
  - [x] Use Ibis to construct an optimized query for one month of data in District 5 (March 2023).
  - [x] Answer Question 1: Are all the data available in District 5 during March 2023? Generate a coverage map.

**Notes For Review**
* Does the methodology work (is the plot is correct)? 
* Are any optimizations possible? The methodology works for one month of data in District 5, but will not work for larger datasets.

**Future Work**
Subsequent PRs will answer the following questions:
* Question 2: If data are missing, do they occur in any spatial or temporal clusters? Do they follow any other spatial characteristics (e.g. urban vs. rural)?
* Question 3: Are there any outliers or unexpected values in the data? 
* Question 4: Are all the data available for all the available districts in recent year, 2023? And again, is there any pattern to the missing data? Are there any odd values?
* Question 5: And what about the most recent decade, 2013-2023?


ian-r-rose:
Thanks for the draft PR @mbobra. I noticed a few things that I think would simplify and speed up the data cleaning and analysis:

First, the process of selecting out the most recent metadata for each sensor ID could be simplified significantly using groupby or window functions. Something along the lines of:

```python
district_5_df.sort_values(["ID", "DATA_VERSION"], ascending=False).group_by("ID").head(1)
```
should do more-or-less the same thing as your logic around looping over the unique IDs.

On `select_data_per_ID`: this is doing a ton of extra work! We can probably bring it from hours down to seconds. A couple of things to note:
1. The return type of the function is just a list of strings for each station ID that has data during the relevant time window. But between all the queries you are pulling effectively the entire dataset for D5 during the time period over the wire. This is probably several gigabytes. Just network transfer may be eating the majority of your time. I recommend only selecting the columns you want, and doing some windowing or aggregation to make it much smaller.
2. You are doing one query for each station ID in D5 for a total of 706 queries. This is 705 more than is necessary, it should be possible to do this in a single query that just returns the relevant station IDs for a given district and time window (this is what's known as the N+1 query problem).
3. It is probably possible to filter on the filename without using a full-on regex, just searching for whether the string contains, e.g., `/d05/` would probably do the trick.

mmarks13 (commented):
I love these plots, Monica. A couple of notes on code optimization:
- Move all processing to snowflake. In this code, `merged_df = out.merge(district_5_recent_version_df, on=['ID'], how='left')`, `out` and `district_5_recent_version_df` are both pandas dataframes. This isn't a huge deal with this data, but will make a difference at scale. 

- Make the plot in one go. Anytime you are using loops and your code is slow, this is usually a good first place to look for optimization. I was able to generate a nearly identical plot in ~2 mins using the code below:

```# Set the aesthetic style of the plots
sns.set_style("whitegrid")

# Create the plot
plt.figure(figsize=(24, 100))
sns.scatterplot(data=merged_df, x='SAMPLE_TIMESTAMP', y='ID',marker = "|", alpha = 0.5)


# Adjust the plot
plt.xticks(rotation=45)
plt.xlabel("Timestamp")
plt.ylabel("ID")
plt.tight_layout()

# Show the plot
plt.show()

mbobra:
@mmarks13 Thanks for these comments -- the speedup on the plot is much appreciated! I addressed all those comments. Let's do the merge optimization in another PR. Please approve if you think this is good to go.

mmarks13 (approved):

PR #28, opened 2024-01-03
Switch to prod s3 bucket and Snowflake environment

### LINKED ISSUE #22: Load historical PeMS data to `RAW_PRD`
We are currently loading to the dev environment, now that #20 is in, we should shift things over to prod. This will probably involve a few steps:

1. Make sure the snowpipe architecture from #20 is deployed to prod
2. Pause pipes
3. Copy all the historical data from the dev bucket to the prod bucket (we don't need to re-run the airflow tasks)
4. Clone the clearinghouse tables in `RAW_DEV` to `RAW_PRD` (we don't need to re-load from the external stage bucket)
5. Unpause pipes

ian-r-rose:
This fixes #22 (more data about the detailed steps taken are over there, the code changes here are quite brief)


melanie-logan (approved):
Looks good!

PR #46, opened 2024-01-24
Data warehouse performance docs

ian-r-rose:
These docs are an adaptation of the docs [here](https://cagov.github.io/data-infrastructure/learning/cloud-data-warehouses/) to Snowflake and PeMS data. The intention is to use them as part of the onboarding/training material for Caltrans staff


mmarks13:
columnar storage image is not rendering. 
image

ian-r-rose:
check again? I pushed a fix for that a couple of minutes ago

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+
+## What is a cloud data warehouse?
+
+Cloud data warehouses (CDWs) are databases which are hosted in the cloud,
+and are typically optimized around analytical queries like aggregations and window functions,
+rather than the typical transactional queries that might support a traditional application.
+This project uses [Snowflake](https://www.snowflake.com/en/) as its data warehouse.
+
+Cloud data warehouses typically have a few advantages over traditional transactional databases for analytical workflows, including:
+
```
mbobra, line 11: It might be useful to describe this as the difference between de-centralized versus centralized, e.g. 

De-centralized (cloud) warehouses have two main advantages over centralized (traditional) databases. Cloud warehouses do not require users to store data locally or serve data. And it is easy to provision additional resources to scale to large datasets.

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+### Constructing queries for cloud data warehouses
+
+With the above understanding of how cloud data warehouses store and process data,
+we can write down a set of recommendations for how to construct efficient queries for large tables stored within them:
+
+* *Only `SELECT` the columns you need.*
+    Columnar storage allows you to ignore the columns you don't need,
+    and avoid the cost of reading it in. `SELECT *` can get expensive!
+* *If the table has a natural ordering, consider setting a partitioning or clustering key.*
```
mbobra, line 70: It might be useful to define the nomenclature (what is the difference between a partitioning and clustering key).

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+
+
+### Primary keys and constraints
+
+A central feature of cloud data warehouses is that storage is separate from compute,
+and data can be processed in parallel by distributed compute resources while they cost money.
+The less communication that needs to happen between these distributed compute resources,
+the faster they can work.
+For this reason, most cloud data warehouses do not support primary keys,
+foreign keys, or other constraints.
```
mbobra, line 98: Again, it might be useful to define primary and foreign keys.

docs/data-warehouse-performance.md
ian-r-rose: I'm not sure I understand your use of decentralized. I'd still consider things like BigQuery and Snowflake to be centralized, just hosted in the cloud vs on-prem, and plenty of traditional databases can also be hosted in the cloud

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+#### Take advantage of partition pruning
+
+There is another problem with the performance of the above queries.
+You can see that there are about two hundred thousand partitions in the entire dataset ("Partitions total").
+And the number of partitions that the queries scanned ("Partitions scanned") was the same!
+That is to say, we needed to read through every single partition,
+even though we were only interested in the partitions corresponding to 2023.
+
+Because we set a clustering key for the `SAMPLE_DATE`, we should be able to efficiently select
+records for specific date ranges.
```
mbobra, line 205: It might be unclear how the user identifies the clustering key.

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+However, query optimizers are not always as good at reasoning about partitions as we would like them to be.
+In this case, Snowflake is unable to reason about the result of the `date_from_parts()` function well enough
+to know what its result will be (even though it will result in a constant date!).
+
+Since Snowflake cannot figure out the result of the `date_from_parts()` call,
+it gives up and scans every partition to figure out if its date is satisfied by the filtering predicates.
```
mbobra, line 211:
```suggestion
it gives up and scans every partition to figure out if its date is satisfied by the filtering predicates. Inefficient queries scan every single partition.
```

docs/data-warehouse-performance.md
mbobra: Sorry, I mixed up the two definitions. I meant centralized vs. de-centralized (not the other way around). I was using it as described in [Robinson+2019](https://arxiv.org/abs/1908.03356):

> In scientific research, which is much more decentralized, the status-quo largely remains a download model, wherein data is moved from central distribution servers to local computing resources. ... To overcome these challenges, scientists need some sort of a big data “platform”, which provides both centralized data storage and a parallel computing framework for analyzing it _in situ_.

I like the idea of contrasting the download model with a cloud-based platform. But this is just a suggestion, please feel free to leave it.

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+you can significantly reduce both execution time and cost for your cloud data warehouses.
+
+### Data layout
+
+Most cloud data warehouses use columnar storage for their data.
+This means that data for each column of a table are stored sequentially in object storage
+(this is in contrast to transactional databases which usually store each row, or record, sequentially in storage).
+[This BigQuery blog post](https://cloud.google.com/blog/topics/developers-practitioners/bigquery-admin-reference-guide-storage) goes into a bit more detail.
+
+![On the left is data laid out in a record-oriented way, where each row's values are contiguous in memory. On the right is data laid out in a columnar way, where each column's values are contiguous in memory.](../images/columnar-storage.png "Columnar storage")
```
britt-allen, line 39:
```suggestion
![On the left is data laid out in a record-oriented way, where each row's values are contiguous in memory. On the right is data laid out in a columnar way, where each column's values are contiguous in memory.](images/columnar-storage.png "Columnar storage")
```

i think this will fix the image. it still doesn't load with the current path

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+
+There are a number of consequences of using columnar storage:
```
britt-allen, line 41:
```suggestion
There are a number of advantages of using columnar storage:
```

consequences has a negative connotation. i'd recommend a more positive word like pros or advantages

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+
+* *Only `SELECT` the columns you need.*
+    Columnar storage allows you to ignore the columns you don't need,
+    and avoid the cost of reading it in. `SELECT *` can get expensive!
+* *If the table has a natural ordering, consider setting a partitioning or clustering key.*
+    For example, if the data in the table consists of events with an associated timestamp,
+    you might want to cluster according to that timestamp.
+    Then events with similar times would be stored near each other in the same or adjacent partitions,
+    and queries selecting for a particular date range would have to scan fewer partitions.
+* *If the table has a partitioning or clustering key already set, try to filter based on that in your queries.*
```
britt-allen, line 75: Might be useful to share how they can find the key of already set

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+    the considerations about partitioning and clustering may seem reminiscent of indexes.
+    Cloud data warehouses usually don't have traditional indexes,
+    but partitioning and clustering keys fill approximately the same role,
+    tailored to the distributed compute model.
+
+
+### Primary keys and constraints
+
+A central feature of cloud data warehouses is that storage is separate from compute,
+and data can be processed in parallel by distributed compute resources while they cost money.
```
britt-allen, line 94: is this an incomplete thought? I don't understand the end of your sentence "while they cost money"

docs/data-warehouse-performance.md
britt-allen: I think people who come from transactional db experience would know these. Not sure if people without that experience need to get bogged down in the deets?

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+!!! note
+    One exception to the above is `NOT NULL` constraints,
+    which can be done cheaply since they don't require information from other tables or partitions to be enforced.
+
+## Interactive Exercise
+
+This exercise is intended to be done live with collaborators.
+It should read fine, but will be more impactful if we set up a lab setting!
+
+We'll be querying PeMS vehicle detectory station data stored in Snowflake.
```
britt-allen, line 118:
```suggestion
We'll be querying PeMS vehicle detector station data stored in Snowflake.
```

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+FROM RAW_PRD.CLEARINGHOUSE.STATION_RAW
+WHERE SAMPLE_DATE > date_from_parts(2023, 1, 1)
+AND SAMPLE_DATE < date_from_parts(2024, 1, 1)
+GROUP BY ID, SAMPLE_DATE
+```
+
+This query took about three minutes to run. Not too bad!
+But when we look at some statistics from the query profile, we start to see some problems:
+
+![Initial query profile](../images/pruning-exercise-initial.png "Initial query profile")
```
britt-allen, line 154:
```suggestion
![Initial query profile](images/pruning-exercise-initial.png "Initial query profile")
```

image not loading that may fix it? or github's preview doesn't work. i recall some markdown magic not working in preview but working when the code is actually merged

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+
+Yikes! We can see from the "Bytes scanned" that this query scanned almost a terabyte of data,
+or about half of the overall dataset.
+Ideally, we should not scan the entire dataset to get some relatively simple statistics for a single year.
+We can also see that from the "Profile Overview", remote disk I/O takes about 86% of the processing time,
+so anything we can do to cut down on disk I/O will likely be worthwhile.
```
britt-allen, line 160: what is disk I/O? might be worth an explanation here

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+GROUP BY ID, SAMPLE_DATE
+```
+
+This ran in under two minutes, and by inspecting the query profile, it's clear why:
+this query scanned a little over half the amount of data as the first one.
+By selecting the only the columns we wanted, we avoided loading a lot of unnecessary data!
+
+![column-pruning](../images/pruning-exercise-column.png "Column pruning")
+
+Using `SELECT *` is often a bad idea in any database,
```
britt-allen, line 193:
```suggestion
Manually selecting all table columns or using `SELECT *` is often a bad idea in any database,
```

Since you didn't technically use select * in the first example

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+
+```sql
+SELECT
+    ID,
+    SAMPLE_DATE,
+    SUM(FLOW_1) AS FLOW
+FROM RAW_PRD.CLEARINGHOUSE.STATION_RAW
+WHERE SAMPLE_DATE > '2023-01-01'
+AND SAMPLE_DATE < '2024-01-01'
+GROUP BY ID, SAMPLE_DATE
```
britt-allen, line 224: why select and group by sample date? the query works without that and likely has better performance

docs/data-warehouse-performance.md
```diff
@@ -0,0 +1,254 @@
...
+This runs in about seventeen seconds, and only scans a small fraction of the partitions in the dataset:
+
+![Partition pruning](../images/pruning-exercise-partition.png "Partition pruning")
+
+You'll also note that the total processing time is much more weighted towards actual data processing
+rather than I/O.
+This is a good thing: we want our compute processes to be working hard for us,
+rather than waiting idly for data to traverse the network while they cost money.
+
+The moral here is: partition pruning can drastically reduce the amount of data you need to process,
```
britt-allen, line 236: is the tl;dr here that partition pruning is filtering your SQL query with a clustering or partition key?

britt-allen (changes requested):
This is incredible, thank you for putting it together! I made a few suggestions and have a few Qs.

docs/data-warehouse-performance.md
ian-r-rose: It's not incomplete, but phrased somewhat awkwardly. I'll try to clarify it. The idea is that the compute resources only cost money while they are processing data, and can shut down when they are not.

docs/data-warehouse-performance.md
ian-r-rose: Yep! It's also useful to use that key for aggregations or windowing in addition to filtering.

docs/data-warehouse-performance.md
ian-r-rose: I'm not sure I understand the question, can you elaborate?

docs/data-warehouse-performance.md
ian-r-rose: Good point

docs/data-warehouse-performance.md
ian-r-rose: Good point, I'll define the term

docs/data-warehouse-performance.md
ian-r-rose: For some reason, mkdocs seems to accept both, but the preview in GitHub needs the latter. I agree removing the `..` is preferable

docs/data-warehouse-performance.md
ian-r-rose: Good idea

docs/data-warehouse-performance.md
britt-allen: I am asking why your query is not like the below. The below achieves the same without selecting sample_date, we don't need that column other than to visually verify our filter works. I am also saying dropping that column likely improves performance a little

```sql
SELECT
    ID,
    SUM(FLOW_1) AS FLOW
FROM RAW_PRD.CLEARINGHOUSE.STATION_RAW
WHERE SAMPLE_DATE > '2023-01-01'
AND SAMPLE_DATE < '2024-01-01'
GROUP BY ID
```

docs/data-warehouse-performance.md
britt-allen: Can we include a TL;DR at the top of this? I personally find it helpful for my learning style to start with a simple understanding then dive into the deets to reinforce. Partition pruning is a technically heavy phrase and this is a tech-heavy doc. I want folks who do and don't have to get bogged down in the deets to have a quick way to understand. Wdyt?

docs/data-warehouse-performance.md
ian-r-rose: I've tried to clarify a bit, how does it look now?

docs/data-warehouse-performance.md
ian-r-rose: This is a different query, and doesn't accomplish the goal of the initial question, which was to compute the total flow per station, *per date*. Your query just computes the total flow per station for the whole year.

britt-allen: oops 😬

ian-r-rose: It is definitely a cheaper query, though :)

docs/data-warehouse-performance.md
ian-r-rose: Done

docs/data-warehouse-performance.md
ian-r-rose: Sure, I've added a note to the top, how does that look?

docs/data-warehouse-performance.md
ian-r-rose: I've added a glossary with some definitions for these terms, does that help? I've also removed references to partition keys: this is a concept in some big data systems, but not in Snowflake.

docs/data-warehouse-performance.md
britt-allen: Thank you! Looks great :)

docs/data-warehouse-performance.md
britt-allen: lol

docs/data-warehouse-performance.md
ian-r-rose: Good catch. I've added a note on how to identify a clustering key that @britt-allen also called out, and added the explicit column name above.

docs/data-warehouse-performance.md
ian-r-rose: added

ian-r-rose:
Thanks for the reviews @mbobra, @britt-allen, and @mmarks13. I *think* I've addressed most of your comments

britt-allen (approved):
Excellent!

PR #51, opened 2024-02-02
creating a staging model for training

britt-allen:


britt-allen:
Can we get this merged @ian-r-rose. The failure is expected in the .yml file. It'll be fixed when trainees reference real values.

transform/models/_models.yml
```diff
@@ -1,14 +1,7 @@
...
-      - name: city
-        description: Their residence
-      - name: license_plate
-        description: Their license plate
-      - name: email
-        description: Their email
+  - name: <model name here>
+    description: <model description here>
+  - name: <model name here>
+    description: <model description here>
```
ian-r-rose, line 7: Should we also add `columns` here?

britt-allen: Done @ian-r-rose

PR #63, opened 2024-02-15
Docs and package updates

britt-allen:


mkdocs.yml
```diff
@@ -26,4 +26,4 @@ nav:
   - Setup: setup.md
   - Writing Documentation: writing-documentation.md
   - Data Warehouse Performance: data-warehouse-performance.md
-  - dbt Project: /dbt_docs/index.html
+  - dbt Project: ../dbt_docs/index.html
```
ian-r-rose, line 29: This might need to be `./` rather than `../`. I suggest trying to build the docs from this branch to test it out

ian-r-rose: That is to say, temporarily add this branch name to here: https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/56eb01a9cae7d92ddfaffa594288698635fcb231/.github/workflows/docs.yml#L5

britt-allen: I forgot how to do that lol, I'll read the docs. Thank you!

ian-r-rose: (`../` might also be correct, I'm not sure off the top of my head)

britt-allen: Found it: https://cagov.github.io/caldata-mdsa-caltrans-pems/writing-documentation/

mkdocs.yml
```diff
@@ -26,4 +26,4 @@ nav:
   - Setup: setup.md
   - Writing Documentation: writing-documentation.md
   - Data Warehouse Performance: data-warehouse-performance.md
-  - dbt Project: /dbt_docs/index.html
+  - dbt Project: dbt_docs_snowflake/index.html
```
ian-r-rose, line 29:
```suggestion
  - dbt Project: dbt_docs/index.html
```

britt-allen: Did this locally before seeing the suggestion here, thanks!

ian-r-rose: I don't see this change in your branch

.github/workflows/docs.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - docs_and_package_updates
```
ian-r-rose, line 6: Docs build looks fixed now, let's remove this and merge!

britt-allen: Oh duh, ty!

PR #92, opened 2024-02-23
Parse clearinghouse status XML

### LINKED ISSUE #89: Parse second column of the station_status table in Clearinghouse
The `CONTENT` column in `RAW_PRD.CLEARINGHOUSE.STATION_STATUS` needs to be further parsed out as it is not human readable. This will allow us to easily get information on station to detector mapping.

ian-r-rose:
Fixes #89 

@kengodleskidot and @britt-allen, can you take a look at the resulting table here to see if it makes sense?

### DIFF
--- a/transform/models/staging/clearinghouse/_clearinghouse.yml
+++ b/transform/models/staging/clearinghouse/_clearinghouse.yml
@@ -2,7 +2,8 @@ version: 2
 
 models:
   - name: stg_clearinghouse__station_meta
-    description: reformating table columns
+    description: |
+      Vehicle detector station metadata from the [PeMS Clearinghouse feeds](https://pems.dot.ca.gov/feeds/clhouse/).
     columns:
       - name: META_DATE
         description: Date of the metadata file.
@@ -39,7 +40,9 @@ models:
       - name: TYPE
         description: Two character string identify the VDS type
   - name: stg_clearinghouse__station_raw
-    description: reformating table columns
+    description: |
+      Vehicle detector station 30-second raw data from the
+      [PeMS Clearinghouse feeds](https://pems.dot.ca.gov/feeds/clhouse/).
     columns:
       - name: ID
         description: Unique VDS (station identifier) value. Use this value to cross-reference with Metadata files.
--- a/transform/models/staging/clearinghouse/_clearinghouse.yml
+++ b/transform/models/staging/clearinghouse/_clearinghouse.yml
@@ -3,9 +3,6 @@ version: 2
 models:
   - name: stg_clearinghouse__station_meta
     description: reformating table columns
-    config:
-      materialized: view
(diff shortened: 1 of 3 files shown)


transform/models/staging/clearinghouse/stg_clearinghouse__station_status.sql
```diff
@@ -0,0 +1,30 @@
...
+    DATE_FROM_PARTS(
+        REGEXP_SUBSTR(STATUS.FILENAME, '{{ date_re }}', 1, 1, '', 1)::INT,
+        REGEXP_SUBSTR(STATUS.FILENAME, '{{ date_re }}', 1, 1, '', 2)::INT,
+        REGEXP_SUBSTR(STATUS.FILENAME, '{{ date_re }}', 1, 1, '', 3)::INT
+    ) AS META_DATE,
+    XMLGET(STATUS.CONTENT, 'station-id'):"$" AS STATION_ID,
+    XMLGET(XMLGET(DETECTOR.VALUE, 'detector'), 'detector-id'):"$"::VARCHAR AS DETECTOR_ID,
+    XMLGET(XMLGET(DETECTOR.VALUE, 'detector'), 'detector-name'):"$"::VARCHAR AS DETECTOR_NAME,
+    XMLGET(XMLGET(DETECTOR.VALUE, 'detector'), 'detector-status'):"$"::VARCHAR AS DETECTOR_STATUS,
+    XMLGET(XMLGET(DETECTOR.VALUE, 'detector'), 'tmdd:last-update-time'):"$"::VARCHAR AS LAST_UPDATE_TIME,
```
britt-allen, line 16: This column is currently empty. Do we know why? Do we need it if it's empty?

britt-allen (approved):
Great work Ian! I have only ever parsed XML in python so nice to see a clean example in SQL.

transform/models/staging/clearinghouse/stg_clearinghouse__station_status.sql
ian-r-rose: I don't know why it's empty. In the spirit of "ELT", and because we don't know that it will *always* be empty, I suggest we keep it here and drop it from downstream models.

transform/models/staging/clearinghouse/_clearinghouse.yml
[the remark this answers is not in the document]
ian-r-rose: Done!

transform/models/staging/clearinghouse/stg_clearinghouse__station_status.sql
britt-allen: Good call, I agree! Merging, but will make sure to bring this PR Ken's attention next time we meet if he hasn't seen it by then.

PR #94, opened 2024-02-23
created seed tables for detector type and diagnostic threshold set testing

britt-allen:


transform/seeds/diag_thresh_set.csv
ian-r-rose, line 1: This file looks like it changes from a CSV to a TSV halfway through?

britt-allen: weird I did not catch that, thank you!

ian-r-rose:
Do we want to think about what schema this seed data should get put in?

ian-r-rose:
Where do these seed tables come from? Can you add documentation for them and what tables they are intended to support? (i.e., are they meant to be joined with `STATION_META`?)

britt-allen:
Hey @kengodleskidot this is a follow up PR to the work we identified needing to be done this past Friday. Can you add your review and let me know if anything is remiss? I also need help defining `DT_NAME` and `DT_VALUE` I could not find descriptions in the docs and links provided.

ian-r-rose (approved):

PR #95, opened 2024-02-24
Model for most recent clearinghouse metadata from each district

### LINKED ISSUE #86: Create dbt model for "most recent station metadata"
The `CLEARINGHOUSE.STATION_META` table contains all of the updates for station metadata for the history of the dataset. In general, a station will show up many times, once for each time it was created or updated.

One useful downstream model would be to create a table for the most recent metadata entry for each station. This would be a useful thing to join in with any of a number of metrics based on VDS data. I have some exploratory work in progress which shouldn't be too hard to get over the line that does this.

ian-r-rose:
The raw source tables for clearinghouse metadata includes the whole history of updates to the metadata. Many analyses will be interested in just the most recent, up-to-date version of the metadata. This creates some intermediate models with that data in it.

Fixes #86


britt-allen (approved):

britt-allen:
Will await Ken's review before merging!

britt-allen:
@kengodleskidot your review is requested on this. @christian-clarion tagging you for awareness

transform/models/intermediate/vds/_vds.yml
```diff
@@ -0,0 +1,163 @@
+version: 2
+
+models:
+  - name: int_vds__station_meta
```
britt-allen, line 4: I think adding `most_recent` to the name like you did down below on line 94 could be helpful

transform/models/intermediate/vds/_vds.yml
```diff
@@ -0,0 +1,163 @@
...
+        description: A descriptive name assocaited with a VDS
+      - name: TYPE
+        description: Two character string identify the VDS type
+  - name: int_vds__most_recent_station_status
+    description: |
+      TMDD Status data from PeMS Clearinghouse. This seems to be the main source of detector
+      ID to station ID mapping.
+
+      This table is restricted to the most recent update for each district, any prior metadata
+      is ignored.
```
britt-allen, line 141: this has the same exact description as the table on line 4, only the name is different. i am confused as they both are restricted to the most recent update according to the descriptions.

transform/models/intermediate/vds/_vds.yml
ian-r-rose: This model is not the "most recent" one, this one has the full history

transform/models/intermediate/vds/_vds.yml
ian-r-rose: Good catch, the one earlier is *not* in fact restricted to the most recent update

transform/models/staging/clearinghouse/stg_clearinghouse__station_status.sql
kengodleskidot, line 1: @ian-r-rose I reviewed the lane number value in the tmdd_meta raw file, compared it to the tables (stg and int) and PeMS. It appears the lane numbering convention in the file is the following: <lane-number>1</lane-number> in the file is lane 1; <lane-number>10</lane-number> in the file is lane 2; <lane-number>100</lane-number> in the file is lane 3; <lane-number>1000</lane-number> in the file is lane 4 and so on. We will want to update the lane numbering convention to count for this nuance.

kengodleskidot (changes requested):
Lane numbering in the "transform/models/staging/clearinghouse/stg_clearinghouse__station_status.sql" file needs to be addressed. See my comment in the file and let me know if you have any questions.

transform/models/staging/clearinghouse/stg_clearinghouse__station_status.sql
ian-r-rose: Thanks, @kengodleskidot. Do you suggest we name these columns `FLOW_1`, `FLOW_10`, `FLOW_100`, etc?

ian-r-rose: Or do you think it would be better to keep the numbering as is and include this in the documentation for the models?

kengodleskidot: I think the use of actual lane numbers (1-8) is what we are looking for if we need to join this data with any other data sets or display the data in a report/visualization. I am not sure the benefit to us or the end user in using lane numbers in multiples of 10 the way the raw data is currently structured. We could keep the numbering convention as is in the stg model and do the lane number conversion in the int table for joining/analysis if needed. If we keep the numbering convention unchanged from the raw data files, documentation on the numbering convention used in this data set would be important to include otherwise a value such as lane number 10 is misleading and lane numbering 100, 1000, etc. are not realistic based on conventional understanding.

ian-r-rose: Okay, in that case I'll revert my most recent commit and we can figure out how to join with the 1, 10, 100 format in downstream models, does that sound good?

kengodleskidot: @ian-r-rose @britt-allen @christian-clarion FYI the vds_config.xml file (https://pems.dot.ca.gov/feeds/vds_config.xml) in combination with the two metadata station related files (https://pems.dot.ca.gov/feeds/clhouse/d03/2024/02/meta/d03_text_meta_2024_02_27.txt and https://pems.dot.ca.gov/feeds/clhouse/d03/2024/02/meta/d03_tmdd_meta_2024_02_27.xml as a recent district 3 example) should form the basis for everything we need related to station information used throughout PeMS. Let's setup a time to discuss how these files can be combined so we can develop all the models needed for the various reports/analysis.

ian-r-rose: @kengodleskidot thanks for pointing the `vds_config.xml`. As far as I can see, this doesn't contain any extra data that isn't in the `STATION_STATUS` and `STATION_META` tables. Do you see anything extra there?

If that's the case, then I'd consider this to be a good summary dataset of the current snapshot of VDS data, but I'd prefer to use the `STATION_STATUS` and `STATION_META` tables since they also include the history of updates to the data. Then the content of this PR would be essentially creating the same summary that `vds_config.xml` has.

kengodleskidot: @ian-r-rose The only additional information I saw in the vds_config.xml file was that it contained the controller to station relationship. I did not see that in the STATION_STATUS or STATION_META tables but if you already have it great! The vds_config.xml also contains all active stations for that date but I believe you already have this, correct?

ian-r-rose: @kengodleskidot I suggest we move discussion of `vds_config.xml` to a new issue, do you have any requests related to the content of this PR?

kengodleskidot: @ian-r-rose Agreed on the vds_config.xml

kengodleskidot (approved):

Issue #84, opened 2024-02-21
PeMS detector health diagnostics: Not enough data samples received


ian-r-rose:
Implement Test number 7 from the [PeMS User Guide](https://pems.dot.ca.gov/Papers/PeMS_Intro_User_Guide_v6.pdf):

<img width="671" alt="image" src="https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/5728311/fff28a0e-f5d0-499f-85ab-9d0b6c65b200">

**Data sources**: 30 second raw data, VDS metadata (controller, station, detector)

It's not super clear to me how this test is different from the second one (cf #79 ), @kengodleskidot do you have any ideas?

britt-allen:
Blocked and related to #79

ian-r-rose:
Why is this blocked? This seems like a fairly straightforward groupby/count.

britt-allen:
@kengodleskidot can add more color here but he essentially said he needs to understand this test more and if/how it is different than test number 2 (#79). So maybe not necessarily blocked, but on hold for now.

britt-allen:
Ken is thinking about changing the logic and result of this test.

kengodleskidot:
On 3/29/24 at 11pm the district 12 traffic data feed went down and no raw data was being sent from the TMC to PeMS. The feed began working and sending raw data to PeMS on 4/3/24. During the time the district 12 traffic feed was down the detector health status for district 12 detectors followed the logic for this test. As a result, I have created a status in the detector_status model that states the district feed is down to clarify what I believe this status represents.

britt-allen:
Can this issue be closed? @kengodleskidot

pingpingxiu-DOT-ca-gov:
@pingpingxiu-DOT-ca-gov  TODO review

Issue #82, opened 2024-02-21
PeMS detector health diagnostics: Flow-occ mismatch


ian-r-rose:
Implement Test number 5 from the [PeMS User Guide](https://pems.dot.ca.gov/Papers/PeMS_Intro_User_Guide_v6.pdf):

<img width="671" alt="image" src="https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/5728311/6829d5e3-23d8-48a6-809e-245a02085e2e">

**Data sources**: 30 second raw data

Issue #80, opened 2024-02-21
PeMS detector health diagnostics: Zero occ or flow


ian-r-rose:
Implement Test number 3 from the [PeMS User Guide](https://pems.dot.ca.gov/Papers/PeMS_Intro_User_Guide_v6.pdf):

<img width="671" alt="image" src="https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/5728311/51a6f07f-1a83-4c8c-acb8-fa7cce76dfef">

**Data sources**: 30 second raw data, VDS metadata (controller, station, detector)

PR #96, opened 2024-02-29
Detector Diagnostic Counts and Aggregations

### LINKED ISSUE #79: PeMS detector health diagnostics: too few data samples
Implement Test number 2 from the [PeMS User Guide](https://pems.dot.ca.gov/Papers/PeMS_Intro_User_Guide_v6.pdf):

image

**Data sources**: 30 second raw data, VDS metadata (controller, station, detector)

kengodleskidot:
This pull request includes files that count and aggregate the 30 second raw data files for detector diagnostic tests included in the PeMS Guide. 
The files also provide the basis to address issue #84 by utilizing the total sample counts to perform the analysis to see if a sufficient number of samples have been received.

Fixes #79, #82, #80.


transform/models/staging/stg_pems__station_30SecRaw.sql
```diff
@@ -0,0 +1,212 @@
...
+
+from {{ source("clearinghouse", "station_raw") }}
+/*
+Currently looks at the same day in previous year. This should be updated to
+look at the previous day once the data refresh brings in more current data
+*/
+where
+    date(SAMPLE_TIMESTAMP) = dateadd(year, -1, current_date())
+    and to_time(SAMPLE_TIMESTAMP) >= '05:00:00'
+    and to_time(SAMPLE_TIMESTAMP) <= '21:59:59'
```
ian-r-rose, line 212: Eventually let's make this configurable so that it's easier to switch to 24 hours. But for now this is good.

transform/models/staging/stg_pems__station_30SecRaw.sql
```diff
@@ -0,0 +1,212 @@
...
+ML, HOV or Ramp).
+*/
+
+from {{ source("clearinghouse", "station_raw") }}
+/*
+Currently looks at the same day in previous year. This should be updated to
+look at the previous day once the data refresh brings in more current data
+*/
+where
+    date(SAMPLE_TIMESTAMP) = dateadd(year, -1, current_date())
```
ian-r-rose, line 210: It will be likely be higher performance to use `SAMPLE_DATE`, as we have set a clustering key for that, allowing for more reliable partition pruning.

This also is where we will want to convert things to use incremental models. The current data refresh schedule is about 1-2 days behind, somewhat deliberately to allow plenty of time for the clearinghouse to be updated.

But for now, I think it should be fine to just switch to `SAMPLE_DATE` and handle the incrementality later.

transform/models/intermediate/int_pems__diagnostic_samples_per_station.SQL
ian-r-rose, line 1: I suggest combining these two models into a single groupby/aggregation. As it is, one model is computing the condition that will be aggregated, and the other is actually doing the aggregation. It will also likely be much faster to replace the case statements with [`COUNT_IF`](https://docs.snowflake.com/en/sql-reference/functions/count_if) (custom case statements will in general be slower than equivalent functions).

So the column` here would look something like

```sql
COUNT_IF(FLOW_1 is not null and OCCUPANCY_1 is not null) as lane1_sample_cnt,
```

I suspect making the above change will be significantly faster and take about half the lines of code.

ian-r-rose (commented):
Cool stuff @kengodleskidot

transform/models/intermediate/int_pems__diagnostic_samples_per_station.SQL
kengodleskidot: This is great and I will work on getting the intermediate file updated using the COUNT_IF statement. Thank you @ian-r-rose!

britt-allen:
@kengodleskidot Ian figured out the group by issue, which was that your CTEs needed two! I added them to the correct places and committed the changes to this branch. See the [latest commit]([b3bece5](https://github.com/cagov/caldata-mdsa-caltrans-pems/pull/96/commits/b3bece534fe1da0fa30d9220cecf13d5d4f2c7e5))

kengodleskidot (commented):
Thanks for your review and updates @britt-allen! Is there anything else you need from me on this PR to move forward?

ian-r-rose (commented):
Sorry @kengodleskidot, I had a review pending for the last few days that I never submitted!

transform/models/staging/detector_health/stg_pems__station_30SecRaw.sql
```diff
@@ -1,4 +1,6 @@
 /*
+This file has been replaced using the COUNT_IF statement that is in the 
+int_pems__diagnostic_samples_per_station.sql. This is being retained for archive purposes.
```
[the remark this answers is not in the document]
kengodleskidot: I am good with deleting this file since it is no longer needed.

transform/models/intermediate/int_pems__diagnostic_samples_per_station.SQL
```diff
@@ -1,57 +1,82 @@
...
-    select * from {{ ref("stg_pems__station_30SecRaw") }}
+    select * from {{ source("clearinghouse", "station_raw") }}
+    /*
+    Currently looks at the same day in previous year. This should be updated to
+    look at the previous day once the data refresh brings in more current data
+    */
+    where
+        DATE(sample_date) = DATEADD(year, -1, CURRENT_DATE())
+        and TO_TIME(sample_timestamp) >= '05:00:00'
+        and TO_TIME(sample_timestamp) <= '21:59:59'
```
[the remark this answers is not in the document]
kengodleskidot: This is currently being done to limit the number of calculations associated with running the file since this process will likely be an incremental model done on a daily basis based on the previous days raw data set. It also allows me to QA/QC this data set against what we have in PeMS (only samples between 5 am and 9:59:59 pm are currently analyzed in PeMS) to ensure the calculations are producing the expected results. Once we implement incremental models or something similar and we look at a complete day's worth of samples I anticipate this code will need to be updated/removed accordingly. I originally had it looking at only the previous days data but that is currently not available in the Snowflake database from my understanding.

transform/models/intermediate/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -0,0 +1,85 @@
...
+        differently the code should be updated as needed to accomodate such a scenario.
+        */
+        COUNT_IF(flow_1 is not null and occupancy_1 is not null) as lane1_sample_cnt,
+        COUNT_IF(flow_2 is not null and occupancy_2 is not null) as lane2_sample_cnt,
+        COUNT_IF(flow_3 is not null and occupancy_3 is not null) as lane3_sample_cnt,
+        COUNT_IF(flow_4 is not null and occupancy_4 is not null) as lane4_sample_cnt,
+        COUNT_IF(flow_5 is not null and occupancy_5 is not null) as lane5_sample_cnt,
+        COUNT_IF(flow_6 is not null and occupancy_6 is not null) as lane6_sample_cnt,
+        COUNT_IF(flow_7 is not null and occupancy_7 is not null) as lane7_sample_cnt,
+        COUNT_IF(flow_8 is not null and occupancy_8 is not null) as lane8_sample_cnt,
```
britt-allen, line 31: Do you know why `lane8_sample_cnt` only has these three distinct values: 0, 1913, or 2031? @kengodleskidot

ian-r-rose: Is it possible there were only three operating sensors in a lane eight for the given day?

kengodleskidot: It may be due to the fact that there are not many freeways in the state system that have 8 lanes. When reviewing the metadata tables I only came across 15 stations with 8 lanes:

select district, id, count(distinct (id)) 
from RAW_DEV.CLEARINGHOUSE.STATION_META
where lanes = 8
group by district, id 

With such a small data set I would not anticipate much variation in the sample counts.

transform/models/intermediate/int_pems__diagnostic_samples_per_station.SQL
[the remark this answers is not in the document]
kengodleskidot: I was able to remove the DATE cast from sample_date but since sample_timestamp includes the date and time I used the TO_TIME cast to extract only the time to do the comparison in the where clause. I tried building the file without to_date on the sample_timestamp and ran into an error.

ian-r-rose:
Ah, right, good point about the time

Ian Rose (he/him)
Principal Data Engineer, Data Services | CA Office of Data and Innovation
(916) 931-8386 | ***@***.***
Follow us: web <https://innovation.ca.gov/> | blog
<https://innovation.ca.gov/blog/> | twitter
<https://twitter.com/californiaodi>

britt-allen (approved):

transform/models/intermediate/detector_health_diag/int_pems__det_diag_high_value_count.sql
```diff
@@ -0,0 +1,87 @@
+with
+source as (
+    select * from {{ source("clearinghouse", "station_raw") }}
```
ian-r-rose, line 3: please select from the staging model rather than directly from the source here:
```suggestion
    select * from {{ ref("stg_clearinghouse__station_raw") }}

```

transform/models/intermediate/detector_health_diag/int_pems__det_diag_high_value_count.sql
```diff
@@ -0,0 +1,87 @@
...
+    hosps.lane5_highocc_cnt,
+    hosps.lane6_highocc_cnt,
+    hosps.lane7_highocc_cnt,
+    hosps.lane8_highocc_cnt
+
+from high_flow_samples_per_station as hfsps
+inner join high_occupancy_samples_per_station as hosps
+    on
+        hfsps.station_id = hosps.station_id
+        and hfsps.sample_date = hosps.sample_date
```
ian-r-rose, line 87: I don't think we should have self joins in these diagnostics. Self joins are expensive and error-prone. All of these metrics are independent of each other, and can be computed in the same CTE.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
ian-r-rose, line 1: In the spirit of reducing the number of expensive joins, I think that `int_pems__diagnostic_samples_per_station` and `int_pems__det_diag_high_value_count` should be consolidated into one SQL model. At the end of the day, I think we should be aiming for a single table for detector status, and the fewer joins we need to do to produce it the better.

As is, these two diagnostic models require *two* passes over the data, and then to do logic about whether a detector is healthy or not, we will have to join them on `sample_date` and `station_id`, which is pretty pricey.

transform/models/staging/detector_health/stg_pems__set_assignment.sql
```diff
@@ -0,0 +1,33 @@
...
+        when district = 11 then 'Urban_D11'
+        when district = 6 then 'D6_Ramps'
+        else 'Urban'
+    end as det_diag_set_id,
+    case
+        when type = 'OR' then 'ramp'
+        else 'mainline'
+    end as det_diag_method_id
+
+from {{ ref('int_vds__station_meta') }}
```
ian-r-rose, line 30: As a matter of practice, we try not to select from intermediate models from staging models (`int_` should always be downstream of `stg_`).

Should this instead be considered an intermediate model?

transform/models/staging/detector_health/stg_pems__set_assignment.sql
```diff
@@ -0,0 +1,33 @@
...
+where
+    DATEADD(year, -1, CURRENT_DATE()) >= _valid_from
+    and DATEADD(year, -1, CURRENT_DATE()) < _valid_to
```
ian-r-rose, line 33: A simpler check is to do `_valid_to` is null. If instead you select from `ref('int_vds__most_recent_station_meta')`, this is done automatically.

transform/models/staging/detector_health/stg_pems__det_diag_threshold_station.sql
ian-r-rose, line 1: Similar to below, this depends on an `int_` model, so shouldn't be considered a `stg_` model.

ian-r-rose (commented):
Thanks @kengodleskidot, great to see this taking shape! I have some suggestions around avoiding unnecessary joins, which I think will be pretty important going forward since this dataset is so large.

transform/models/staging/detector_health/stg_pems__set_assignment.sql
kengodleskidot: Thanks for catching this and I will update the reference!

kengodleskidot (commented):
Update reference in stg_pems__set_assignment.sql and look into combining the staging detector diagnostics files into a int_ file using CTE's

ian-r-rose:
> Before I ran into the ymllint issue I was thinking of combining stg_pems__set_assignment and stg_pems__det_diag_threshold_station as CTE's in a single int_ file. Does that seem like a worthwhile effort based on your review?

Sure, I tend to like having fewer models rather than more, especially if you don't envision them being used separately

transform/models/staging/detector_health/_detector_health.yml
```diff
@@ -0,0 +1,69 @@
+version: 2
+
+models:
+  - name: stg_pems__set_assignment
```
ian-r-rose, line 4: This model no longer exists, since you've made it into a CTE. I think we can remove it here.

transform/models/staging/detector_health/_detector_health.yml
```diff
@@ -0,0 +1,69 @@
...
+        description: Two character string identify the VDS type
+      - name: DET_DIAG_SET_ID
+        description: |
+          Assigns 1 of 5 values for detector diagnostic evaluations based
+          on station district
+      - name: DET_DIAG_METHOD_ID
+        description: |
+          Assigns 1 of 2 values (mainline or ramp) for detector diagnostic
+          evaluations based on station type
+  - name: stg_pems__det_diag_threshold_station
```
ian-r-rose, line 33: You've moved this model into the `intermediate` directory, so let's move the YAML docs for it there as well

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -0,0 +1,118 @@
...
+        COALESCE(lane2_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane2_too_few_samples,
+        COALESCE(lane3_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane3_too_few_samples,
+        COALESCE(lane4_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane4_too_few_samples,
+        COALESCE(lane5_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane5_too_few_samples,
+        COALESCE(lane6_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane6_too_few_samples,
+        COALESCE(lane7_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane7_too_few_samples,
+        COALESCE(lane8_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane8_too_few_samples
+
+    from samples_per_station
+    group by all
```
ian-r-rose, line 107: `samples_per_station` is already aggregated by `sample_date` and `station_id`, so this `group by all` doesn't really do what we want. Instead, this also groups by `sample_date`, `station_id`, and all of the `*_too_few_samples` columns.

I suggest instead doing a `select samples_per_station.*`, including your new columns, and not aggregating here.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -0,0 +1,118 @@
...
+)
+
+select
+    sps.*,
+    ddtfs.* exclude (sample_date, station_id)
+
+from samples_per_station as sps
+left join det_diag_too_few_samples as ddtfs
```
ian-r-rose, line 115: These two table have the same source table (`samples_per_station`) and don't need to be joined (which is expensive). Instead you can just continue adding the columns you need in the CTEs.

transform/models/intermediate/detector_health_diag/int_pems__det_diag_high_value_count.sql
```diff
@@ -0,0 +1,135 @@
...
+        on
+            ddsa.det_diag_set_id = dtv.dt_set_id
+            and ddsa.det_diag_method_id = dtv.dt_method
+),
+
+/*
+This SQL file counts the number of flow and occupancy values that exceed
+detector threshold values for a station based on the station set assignment.
+*/
+high_flow_samples_per_station as (
```
ian-r-rose, line 69: This CTE and `high_occupancy_samples_per_station` should be combined into one to avoid needing the inner join bringing them together (which is expensive).

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -0,0 +1,118 @@
...
+source as (
+    select * from {{ source("clearinghouse", "station_raw") }}
+    {% if is_incremental() %}
+        where sample_date > (
+            select dateadd(day, -2, max(sample_date)) 
+            and TO_TIME(sample_timestamp) >= '05:00:00'
+            and TO_TIME(sample_timestamp) <= '21:59:59'            
+            from {{ this }}
+        )
+    {% endif %}
```
ian-r-rose, line 17: Ooh, incrementality!

ian-r-rose (commented):
Thanks @kengodleskidot, this is looking great! I have a few comments around organization and avoiding unnecessary joins, but I think this is getting close

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
britt-allen: ;)

transform/models/intermediate/detector_health_diag/int_pems__det_diag_high_value_count.sql
britt-allen: oops my current commit does not address this, I was worried about the `and ddsa.dt_name = 'high_occ'` and the `and ddsa.dt_name = 'high_flow'` filtering out each other if that makes sense, but i suppose that will not happen

britt-allen: I'll address this tomorrow!

ian-r-rose: Sounds good! I don't fully understand the `'high_occ'` and `'high_flow'` conditions, but the join condition with them looks odd to me (usually a join condition involves both tables). It may be that you want to move them into the `COUNT_IF` logic

britt-allen:
commented out incrementality for now until I resolve #120

britt-allen:
I didn't believe green check marks existed anymore and then I saw two 🥲 @ian-r-rose @kengodleskidot

britt-allen (commented):
I think this all finally looks good! 🤞🏾

ian-r-rose (commented):
Thanks @britt-allen and @kengodleskidot, this is looking great! I have just one minor comment, but the code is looking just about ready to go in to me

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -0,0 +1,83 @@
...
+
+with
+source as (
+    select * from {{ ref ('stg_clearinghouse__station_raw') }}
+    where TO_TIME(sample_timestamp) >= '05:00:00' and TO_TIME(sample_timestamp) <= '21:59:59'
+    {% if is_incremental() %}
+        -- Look back two days to account for any late-arriving data
+        where sample_date > (
+            select dateadd(day, -2, max(sample_date)) from {{ this }}
+        )
```
ian-r-rose, line 16: This is failing CI because when `is_incremental()` evaluates to `True` it gets two `where` clauses. Instead, the block inside `is_incremental()` should just be an `and` clause.

kengodleskidot: @ian-r-rose would this error result in me being unable to create the table? I was updating all of my models to account for the data unpivot but when I go to run this model I had to cancel it since it was running too long. The code is looking good @britt-allen I just wanted to do some QA/QC on the results.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -0,0 +1,83 @@
...
+        -- # of samples < 60% of the max collected samples during the test period
+        -- max value: 2 samples per minute times 60 mins/hr times 17 hours in a day which == 1224
+        -- btwn 1 and 1224 is too few samples
+        COALESCE(sample_ct between 1 and (0.6 * (2 * 60 * 17)), false) as too_few_samples
+
+    from samples_per_station
+)
+
+select * from det_diag_too_few_samples
+order by station_id desc
```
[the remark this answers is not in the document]
britt-allen: What cluster_by directive are you referring to? Also, am I understanding correctly that you're saying I am messing up column order? If not, can oyu expand further, please?

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
britt-allen: I agree, I could benefit from another walk through, my grasp of these concepts is shaky at best. @kengodleskidot Ian will join our call this Wednesday at 3:30 pm so we can talk through this.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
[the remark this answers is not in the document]
ian-r-rose: I'm referring to the `cluster_by` in the incremental config. This indicates that the final model will be ordered by `sample_date`, rather than `station_id`. So this is in conflict with that, and sets it up for additional re-sorting by the snowflake clustering service (which costs money).

I think it would be okay to do `order by sample_date, station_id`, or to just omit it.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
britt-allen: good catch of the double `where` clauses btw, thank you!

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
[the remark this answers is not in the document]
britt-allen: Makes sense, thank you!

britt-allen:
Alrighty, this is ready, who wants to do the honors? @kengodleskidot @ian-r-rose

ian-r-rose (approved):
Looks good, nice work @kengodleskidot and @britt-allen!

PR #100, opened 2024-03-01
Caltrans PeMS: Exploratory Data Analysis

mbobra:
Added documentation titled "Caltrans PeMS: Exploratory Data Analysis".

cc @kimberlyhicksodi @eaeslami


mmarks13 (commented):
Looks great! Thanks for pulling this together. 

The only thing I notice is the figure captions rendering on the same line as the figures. I think there needs to be an additional carriage return between the figure and the caption. 

`Figure 1. An example of a null observation.
Figure 1. An example of a null observation.

Figure 2. An example of an observation containing all zero values.
Figure 2. An example of an observation containing all zero values.

Figure 3. Average lane 1 flow from all California sensors in 2022 and 2023. Black dots represent sensors with all null values.
Figure 3. Average lane 1 flow from all California sensors in 2022 and 2023. Black dots represent sensors with all null values.

Figure 4. Sensors reporting all null values for 2022 and 2023.
Figure 4. Sensors reporting all null values for 2022 and 2023.

Figure 5. Sensors reporting all zero values for 2022 and 2023. These appear to be centered around the major metropolitan areas.
Figure 5. Sensors reporting all zero values for 2022 and 2023. These appear to be centered around the major metropolitan areas.

Figure 6. The speed (left) and flow (right) of traffic over the Bay Bridge on Tuesday, September 13, 2022.
Figure 6. The speed (left) and flow (right) of traffic over the Bay Bridge on Tuesday, September 13, 2022.`

mmarks13 (approved):
Looks great! Thanks Monica!

eaeslami:
@mbobra would it be possible to link to the markdown pages when you make these kinds of edits so that we can more easily see the work that's being referenced?

Issue #76, opened 2024-02-21
Transform raw VDS XML files into structured data


ian-r-rose:
30 second samples seem to land as individual XML files which have flow, count, and occupancy values for each detector. We'll want to develop a pipeline to roll those samples into structured timeseries parquet files.

ian-r-rose:
@pingpingxiu-DOT-ca-gov can you update here with what you have been learning about the internal caltrans ETL pipelines?

ian-r-rose:
Similar to #77, we are having to re-think the data access story for the VDS data. Closing this issue as not planned.

PR #102, opened 2024-03-05
Work in progress parsing individual district VDS sample xml files

ian-r-rose:
This is the start of a script for parsing individual XML VDS sample files, fixing #76. It currently targets the D7 sample files that have been loaded to S3, but will need to be extended to the other district files as they become available.

cc @pingpingxiu-DOT-ca-gov for visibility

### DIFF
--- a/dags/parse_d7_vds_data.py
+++ b/dags/parse_d7_vds_data.py
@@ -1,0 +1,75 @@
+import glob
+import os
+import tarfile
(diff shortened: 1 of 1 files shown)


ian-r-rose:
Since we are likely not going this route for getting VDS data into the data warehouse, I'm closing this PR. Should things change, we can always reopen

PR #106, opened 2024-03-11
disabled trailing whitespaces 

britt-allen:
(From Ian: This rule has slightly different semantics from the prettier yaml formatter. Disable in favor of the latter.)

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -8,3 +8,4 @@ rules:
     required: false
   truthy:
     allowed-values: ["true", "false"]
+trailing-spaces: disable


ian-r-rose (approved):
Thanks @britt-allen

britt-allen:
@kengodleskidot this should fix the trailing whitespace issue you're currently running into on your PR. From @ian-r-rose "It looks like prettier will trim trailing whitespace most of the time, but not in multi-line text blocks. This makes a certain amount of sense to me, as those are intended to be literal text, rather than configuration/code." So in other words prettier (lint/fix) was not "fixing" trailing whitespaces that were literal whitespaces in multi-line text

britt-allen:
I misspoke, this would not have fixed the trailing whitespace issue you were running into on PR #96. There was a real trailing whitespace in a SQL file that needed to be addressed. I fixed it in this commit: https://github.com/cagov/caldata-mdsa-caltrans-pems/pull/96/commits/188bc794dae5acec33587a76354d80f8309172ad @kengodleskidot

PR #112, opened 2024-03-13
code to select samples that are between 1 and 1224, which constitutes as too few samples for a given lane

### LINKED ISSUE #79: PeMS detector health diagnostics: too few data samples

britt-allen:
This is still in progress. I need to:

- [x] Figure out how to include lane 7 and 8 even if no sample meet the criteria. Currently those two lane pass the check, but the way the query is constructed it causes the check to not be run against the other lanes.
- [x] Materialize this as incremental

Closes #79

### DIFF
--- a/transform/models/intermediate/int_pems__too_few_samples_test.sql
+++ b/transform/models/intermediate/int_pems__too_few_samples_test.sql
@@ -1,0 +1,85 @@
+-- # of samples < 60% of the max collected samples during the test period.
+-- max value is a static value, which should be two samples per minute times 60 mins/hr times 17 hours in a day 
+-- (1224 is the static number of samples) - 
+-- this max value doesnt need to be calculated at the station_id level bc it's the same
+-- btwn 1 and 1224 is too few samples.
+
+with
+diagnostic_samples_per_station as (
+    select * from {{ ref("int_pems__diagnostic_samples_per_station") }}
+),
+
+lane1_too_few_samples as (
+    select lane1_sample_cnt
+
+    from diagnostic_samples_per_station
+    where lane1_sample_cnt between 1 and (0.6 * (2 * 60 * 17))
+),
+
+lane2_too_few_samples as (
+    select lane2_sample_cnt
+
+    from diagnostic_samples_per_station
+    where lane2_sample_cnt between 1 and (0.6 * (2 * 60 * 17))
+),
+
+lane3_too_few_samples as (
+    select lane3_sample_cnt
+
+    from diagnostic_samples_per_station
+    where lane3_sample_cnt between 1 and (0.6 * (2 * 60 * 17))
+),
+
+lane4_too_few_samples as (
+    select lane4_sample_cnt
+
+    from diagnostic_samples_per_station
+    where lane4_sample_cnt between 1 and (0.6 * (2 * 60 * 17))
+),
+
+lane5_too_few_samples as (
+    select lane5_sample_cnt
+
+    from diagnostic_samples_per_station
+    where lane5_sample_cnt between 1 and (0.6 * (2 * 60 * 17))
+),
+
+lane6_too_few_samples as (
+    select lane6_sample_cnt
+
+    from diagnostic_samples_per_station
+    where lane6_sample_cnt between 1 and (0.6 * (2 * 60 * 17))
+)
+
+-- , lane7_too_few_samples as (
+--     select lane7_sample_cnt
+
+--     from diagnostic_samples_per_station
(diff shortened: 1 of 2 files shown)


transform/models/intermediate/int_pems__too_few_samples_test.sql
```diff
@@ -0,0 +1,85 @@
...
+-- , lane8_too_few_samples.*  -- currently lane 7 and lane 8 causes the query to produce no results!
+
+from lane1_too_few_samples,
+    lane2_too_few_samples,
+    lane3_too_few_samples,
+    lane4_too_few_samples,
+    lane5_too_few_samples,
+    lane6_too_few_samples
+-- , lane7_too_few_samples 
+-- , lane8_too_few_samples -- currently lane 7 and lane 8 causes the query to produce no results!
```
ian-r-rose, line 85: This causing the query to produce no results is indicative of the query logic being wrong. Each of the lanes in this calculation *should* be entirely independent of the others, so if including more makes the number of *rows* returned different, it means that something is amiss.

What you are doing here is a natural join of _all_ of the CTEs you've constructed above. This is both expensive (as all joins are), and not what you want.

I recommend ditching most of the CTEs here. Each CTE is effectively a new table from a semantic perspective. We don't want a bunch of tables, we want *one* table with a bunch of columns.

britt-allen: Thank you for this feedback Ian!

transform/models/intermediate/int_pems__too_few_samples_test.sql
```diff
@@ -0,0 +1,34 @@
...
+    coalesce(lane3_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane3_too_few_samples,
+    coalesce(lane4_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane4_too_few_samples,
+    coalesce(lane5_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane5_too_few_samples,
+    coalesce(lane6_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane6_too_few_samples,
+    coalesce(lane7_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane7_too_few_samples,
+    coalesce(lane8_sample_cnt between 1 and (0.6 * (2 * 60 * 17)), false) as lane8_too_few_samples
+
+    -- # of samples < 60% of the max collected samples during the test period
+    -- max value: 2 samples per minute times 60 mins/hr times 17 hours in a day which == 1224 
+    -- btwn 1 and 1224 is too few samples
```
ian-r-rose, line 31: Nit: can you comment before the relevant lines, rather than after?

transform/models/intermediate/int_pems__too_few_samples_test.sql
```diff
@@ -0,0 +1,34 @@
...
+    )
+}}
+
+with
+diagnostic_samples_per_station as (
+    select * from {{ ref("int_pems__diagnostic_samples_per_station") }}
+    {% if is_incremental() %}
+        where sample_date > (
+            select dateadd(day, -2, max(sample_date)) from {{ this }}
+        )
```
ian-r-rose, line 13: This won't really work until the model you are selecting from also has the full date range (right now it just has one day). But I think it's fine to have here until then

ian-r-rose (commented):
Similar to my review in #96, I think we should be aiming to have all of these diagnostics in the same table, rather than spread out across many models.

I think this is fine to merge, but as some follow-up work we should work to consolidate these checks

transform/models/intermediate/int_pems__too_few_samples_test.sql
britt-allen: Oui!

transform/models/intermediate/int_pems__too_few_samples_test.sql
britt-allen: Thank you for that additional insight!

britt-allen:
> Similar to my review in #96, I think we should be aiming to have all of these diagnostics in the same table, rather than spread out across many models.
> 
> I think this is fine to merge, but as some follow-up work we should work to consolidate these checks

I think I understand this and based on that I agree. Let's discuss more at 1:30!

PR #118, opened 2024-03-14
5-minute aggregated VDS data

### LINKED ISSUE #117: Aggregate raw 30-second data to five minutes
Many downstream metrics (e.g., speed) are based on five-minute aggregations of the raw VDS data. Ideally, this would include imputations, but for the time being, we want to aggregate the un-imputed raw data to 5 minutes to unblock further metric development.

ian-r-rose:
This creates incremental model aggregating 30 second VDS data to five minutes. There is no attempt to do any imputation, just rolling up the sample data to a larger time bucket

Fixes #117 

cc @AzamBeg-Caltrans

### DIFF
--- a/transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
+++ b/transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
@@ -50,16 +50,7 @@ aggregated as (
         avg(occupancy_5) as occupancy_5,
         avg(occupancy_6) as occupancy_6,
         avg(occupancy_7) as occupancy_7,
-        avg(occupancy_8) as occupancy_8,
-        -- Average of all the speed values
-        avg(speed_1) as speed_1,
-        avg(speed_2) as speed_2,
-        avg(speed_3) as speed_3,
-        avg(speed_4) as speed_4,
-        avg(speed_5) as speed_5,
-        avg(speed_6) as speed_6,
-        avg(speed_7) as speed_7,
-        avg(speed_8) as speed_8
+        avg(occupancy_8) as occupancy_8
     from station_raw
     group by id, sample_date, sample_timestamp_trunc
 )
--- a/transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
+++ b/transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
@@ -1,0 +1,66 @@
+{{ config(
+    materialized="incremental",
+    cluster_by=["sample_date"],
+    unique_key=["ID", "SAMPLE_TIMESTAMP"],
+) }}
+
+with station_raw as (
+    select
+        *,
+        /* Create a timestamp truncated down to the nearest five
+         minute bucket. This will be the the timestamp on which
+         we aggregate. If a 30-second interval straddles two different
+         buckets, it will be assigned to the one latter one due to
+         the floor() call.
+        */
+        dateadd(
+            'minute',
+            floor(minute(sample_timestamp) / 5) * 5,
+            trunc(sample_timestamp, 'hour')
+        ) as sample_timestamp_trunc
+    from {{ ref('stg_clearinghouse__station_raw') }}
+    {% if is_incremental() %}
+        -- Look back two days to account for any late-arriving data
+        where sample_date > (
+            select dateadd(day, -1, max(sample_date)) from {{ this }}
+        )
+    {% endif %}
+),
+
+aggregated as (
+    select
+        id,
+        sample_date,
+        sample_timestamp_trunc as sample_timestamp,
+        -- Sum of all the flow values
+        sum(flow_1) as flow_1,
+        sum(flow_2) as flow_2,
+        sum(flow_3) as flow_3,
+        sum(flow_4) as flow_4,
+        sum(flow_5) as flow_5,
+        sum(flow_6) as flow_6,
+        sum(flow_7) as flow_7,
+        sum(flow_8) as flow_8,
+        -- Average of all the occupancy values
+        avg(occupancy_1) as occupancy_1,
+        avg(occupancy_2) as occupancy_2,
+        avg(occupancy_3) as occupancy_3,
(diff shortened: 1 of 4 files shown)


ian-r-rose:
@AzamBeg-Caltrans this is ready for review. If you'd like to inspect what the table will look like, it's built in CI as `TRANSFORM_DEV.DBT_CLOUD_PR_15_115_VDS.INT_VDS__FIVE_MINUTE_STATION_AGG`.

transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
```diff
@@ -0,0 +1,67 @@
...
+        -- Average of all the occupancy values
+        avg(occupancy_1) as occupancy_1,
+        avg(occupancy_2) as occupancy_2,
+        avg(occupancy_3) as occupancy_3,
+        avg(occupancy_4) as occupancy_4,
+        avg(occupancy_5) as occupancy_5,
+        avg(occupancy_6) as occupancy_6,
+        avg(occupancy_7) as occupancy_7,
+        avg(occupancy_8) as occupancy_8,
+        -- Average of all the speed values
```
AzamBeg-Caltrans, line 54: The speeds will be calculated by using flow and occupancy. I will try to figure this out in the same branch.

ian-r-rose: Do you know where these speed values are coming from? They are in the raw clearinghouse files.

I'd recommend creating a new branch to run your experiments, rather than using this one.

AzamBeg-Caltrans: Current speed columns are coming from PeMS; how speeds are calculated is not declared by the vendor. We want to calculate our own speeds.

ian-r-rose: Okay, that's fine. We can ignore these values.

ian-r-rose: Would you rather ignore them downstream, or drop them from this aggregation all-together?

AzamBeg-Caltrans: In my opinion, we can drop these from aggregation altogether. We can verify with Ken and Zhenyu.

ian-r-rose: Okay!

AzamBeg-Caltrans: Here is the Python code snippet I had used some time ago for calculating speed:

    for ln in range(1, lane_cnt+1):
        df1[f'SPD{ln}'] =  (df1[f'LANE_{ln}_FLOW'] + df1[f'LANE_{ln}_TRUCK_FLOW'] )   \
                          / df1[f'LANE_{ln}_OCC'] * df1[f'LANE_{ln}_G'] / 600 

Although df1[f'LANE_{ln}_G'] varies slightly with lane number, it can be assumed as 25, until we finalize the imputation method. 

It had seemed to me that between ~10 PM and ~4 AM, 'free-flow speeds' were assigned instead of calculated values. 
The speeds are listed in the 'Speeds and G-factors' section of the PeMS help page:
     https://pems.dot.ca.gov/?dnode=Help&content=help_calc#speeds

transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
```diff
@@ -0,0 +1,58 @@
...
+         the floor() call.
+        */
+        dateadd(
+            'minute',
+            floor(minute(sample_timestamp) / 5) * 5,
+            trunc(sample_timestamp, 'hour')
+        ) as sample_timestamp_trunc
+    from {{ ref('stg_clearinghouse__station_raw') }}
+    {% if is_incremental() %}
+        -- Look back two days to account for any late-arriving data
```
britt-allen, line 24: did you mean for this to be "Look back one day..." since you're looking back 1 day below?

transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
```diff
@@ -0,0 +1,58 @@
...
+        where sample_date > (
+            select dateadd(day, -1, max(sample_date)) from {{ this }}
```
britt-allen, line 26:
```suggestion
            select dateadd(day, -2, max(sample_date)) from {{ this }}
```

if this is indeed meant to be two days

britt-allen (approved):
This looks great! I love the macro that allows for dynamically changing the warehouse.

transform/models/intermediate/vds/int_vds__five_minute_station_agg.sql
ian-r-rose: Good catch, fixed now

kengodleskidot:
Looks good @ian-r-rose! We are missing some stations (for 10/01/2020 station count in 5-min summary PeMS is 18459 and 5 minute-agg-table in Snowflake is 16019) but the aggregations dates/times, flows and occupancy values look reasonable. I will spot check some station data where we have 288 samples in a day for QA/QC and let you know what I find.
image
image

ian-r-rose:
Thanks for looking @kengodleskidot. I wonder what the missing stations are? Are they from any particular district or region?

ian-r-rose:
I think we probably have enough to move forward with this aggregation to unblock some of the downstream metrics work, however. Do you agree @kengodleskidot?

kengodleskidot:
@ian-r-rose I agree that we can move forward so we can perform calculation on the downstream metrics. Getting the speed calculated based on the flow and occupancy numbers is the next step I see so that we can perform these downstream metric calculations. We do have the station count discrepancy that we ultimately need to figure out but it should not hold us up.

ian-r-rose:
Created a new issue for the station counts here, will move forwards with this aggregated data for now. Thanks all!

PR #124, opened 2024-03-21
Unpivot VDS Data

### LINKED ISSUE #121: Unpivot VDS data to long format
The VDS data are currently stored in a "wide" format. This means that each lane is stored in its own column, i.e.

| ID | Date | Timestamp | flow_1 | flow_2 | ... | flow_8 | occ_1 | occ_2 | ... | occ_8 |
|---|------|------------|-------|--------|----|-----|--|--|--|--|
| | | | | | | | | | | |

This is typically considered worse for data analytics for a few reasons:

1. Relevant data is stored in the column name
2. It's more cumbersome to do aggregations (look at how many things are repeated for each lane in, e.g. #96 )
3. We carry around columns for station IDs that don't need them: most stations don't have eight lanes of detectors

I think it would be a good idea to standardize on "long" format:

| ID | Date | Timestamp | Lane | Flow | Occupancy |
|--|--|--|--|--| -- |

That is to say, we store the lane number in a column, and the metrics are not in repeated columns.

If we do this, it would also be a natural point to introduce detector IDs by matching on the lane number (which @kengodleskidot brought up in #95 )

ian-r-rose:
Fixes #121 

Still a work in progress, but I wanted to call out how `int_vds__five_minute_station_agg.sql` is simplified with the unpivoted data.

I've also done some minor reorganization to keep clearinghouse data in `clearinghouse` subdirectories. Still mulling over some of the directory organization here.


ian-r-rose:
Okay, I think this is ready for review. There is some renaming/restructuring that can be mostly ignored, I suggest focusing on `int_clearinghouse__five_minute_station_agg` and `stg_clearinghouse__station_raw`

britt-allen:
Thanks for doing this! I thought we discussed getting in #96 first before we did this work?

ian-r-rose:
> Thanks for doing this! I thought we discussed getting in https://github.com/cagov/caldata-mdsa-caltrans-pems/pull/96 first before we did this work?

That wasn't my understanding. This can go in before or after #96, but either way the aggregations there will have to be unpivoted.

britt-allen (approved):
This all looks really great, thanks for flattening the data and for the cleanup/organization.

PR #127, opened 2024-03-22
Preliminary speed equation that still needs refinement

kengodleskidot:
I commented out the incremental code so the model only processes one year's worth of data for testing and cost purposes.


transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -37,7 +39,65 @@ aggregated as (
...
+        *,
+        --For speed I used the following formula to get speed: 
+        --SPEED = SUM(FLOW)/AVG(OCCUPANCY)/600 which resulted in 
+        --values from 0 to 5
+        --On 3/22/24 I updated the formula to use a vehicle effective length
+        --of 22 feet (16 ft vehicle + 6 ft detector zone) feet and using 
+        --a conversion to get miles per hour (5280 ft / mile and 12
+        --5-minute intervals in an hour).
+        case
+            when flow_1 = 0 or occupancy_1 = 0 then 0
```
ian-r-rose, line 57: Sorry to shift things out from under you @kengodleskidot, now with the unpivoted data we no longer have `flow_1`, `flow_2`, etc, it's just `flow` and the lane number is in the `lane` column

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -37,7 +39,65 @@ aggregated as (
...
+            when flow_1 is null or occupancy_1 is null then null
+            --else (flow_1 * 22) / occupancy_1 / 5280 * 12
```
ian-r-rose, line 59: Neat! I'm thinking through this formula to make sure it makes sense to me, but I'm liking this approach more than the original formula.

ian-r-rose (commented):
Nice start!

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -38,6 +38,24 @@ aggregated as (
...
+        --values from 0 to 5
+        --On 3/22/24 I updated the formula to use a vehicle effective length
+        --of 22 feet (16 ft vehicle + 6 ft detector zone) feet and using 
+        --a conversion to get miles per hour (5280 ft / mile and 12
+        --5-minute intervals in an hour).
+        case
+            when volume = 0 or occupancy = 0 then 0
+            when volume is null or occupancy is null then null
+            else (volume * 22) / occupancy * (1 / 5280) * 12
+        end as speed
```
ian-r-rose, line 57: A few notes to make this logic a bit simpler and more efficient to calculate:

1. We don't need to check when `volume=0`, since the formula already evaluates to zero if volume is 0.
2. When `null` is involved in an arithmetic operation, the result is always `null`, so the second check isn't needed.
3. We can use the convenient `nullifzero` function to help with the `occupancy=0` check.

Putting those all together, we can replace this somewhat complicated case statement with a simpler formula that should also be faster:

```suggestion
        (volume * 22) / nullifzero(occupancy) * (1 / 5280) * 12 as speed
```

ian-r-rose:
@kengodleskidot I've gone ahead and dropped the current 5-min-agg table (which will need to be rebuilt from scratch anyways), so hopefully CI will be a bit smoother now.

ian-r-rose (approved):
Looks good to me, thanks @kengodleskidot!

transform/models/intermediate/clearinghouse/_clearinghouse.yml
```diff
@@ -173,3 +173,8 @@ models:
         description: The sum of the flow values for a detector over the sample period.
       - name: OCCUPANCY
         description: The average of the occupancy values over the sample period.
+      - name: SPEED
+        description: |
+          Preliminary speed calculation in miles/hour based on the simplified version
+          of the speed formula located at
+          https://pems.dot.ca.gov/Papers/vanzwet_gfactor.pdf
```
britt-allen, line 180: love this linking back!

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -38,6 +43,20 @@ aggregated as (
...
+aggregated_speed as (
+    select
+        *,
+        --For speed I used the following formula to get speed:
+        --SPEED = SUM(FLOW)/AVG(OCCUPANCY)/600 which resulted in
+        --values from 0 to 5
+        --On 3/22/24 I updated the formula to use a vehicle effective length
+        --of 22 feet (16 ft vehicle + 6 ft detector zone) feet and using
+        --a conversion to get miles per hour (5280 ft / mile and 12
+        --5-minute intervals in an hour).
```
britt-allen, line 57: I really appreciate your thoughtfulness in explaining your logic for other collaborators.

britt-allen (commented):
Apologies for the late review, I know this is merged already but I still wanted to check out your work!

PR #130, opened 2024-03-25
Don't do "relation is none" check

### LINKED ISSUE #129: `get_snowflake_refresh_warehouse()` uses large warehouse when it shouldn't
When running in incremental mode, a [CI job](https://california.getdbt.com/deploy/6/projects/10/runs/2235) chose the large refresh warehouse, when the intention was that large warehouses were only run in two circumstances:

1. Upon first build of a table
2. When the `--full-refresh` flag is set

I think the first check is not working as intended, probably because `relation is none` in the compile phase of the macro logic. This is because "no sql is executed in the compile phase" (docs [here](https://docs.getdbt.com/reference/dbt-jinja-functions/execute)).

If that's the case, I can't think of a way to reliably handle the first case, and it might be best to just turn it off, and require `--full-refresh=True` to get the truly large warehouse. This is a bit less clever, but cuts down on the possibility of accidentally choosing the big one and spending a lot of money unnecessarily.

ian-r-rose:
The "relation is none" check in our `get_snowflake_refresh_warehouse() appears to not work during the compile phase.

See #129 for more info. Fixes #129

### DIFF
--- a/transform/macros/get_snowflake_refresh_warehouse.sql
+++ b/transform/macros/get_snowflake_refresh_warehouse.sql
@@ -5,7 +5,7 @@
   {% else %}
     {% set suffix = 'DEV' %}
   {% endif %}
-  {% if flags.FULL_REFRESH or relation is none %}
+  {% if flags.FULL_REFRESH %}
     {% set size = '4XL' %}
   {% else %}
     {% set size = 'XS' %}


britt-allen (approved):
Thank you for this quick and important fix

PR #132, opened 2024-03-25
Put new detector diagnostics in detector health schema

ian-r-rose:
Put these in the `DETECTOR_HEALTH` schema, similar to #94


britt-allen (approved):
I knew I was missing something, thank you @ian-r-rose!

britt-allen (approved):

PR #134, opened 2024-03-25
Disabling this model until we resolve a faulty join that causes it to blow up

### LINKED ISSUE #133: Disable `int_pems__diagnostic_samples_per_station` model
Currently this model's rows are being blown up because of a faulty join. This is causing builds to not build and is expensive. We're disabling the model until we can resolve the issue with the join as seen in the Snowflake query profile screenshot below. On the left the join happening in the `int_pems__det_diag_set_assignment` model is duplicating rows nearly 6x.
image

britt-allen:
fixes #133


ian-r-rose (approved):
Thanks @britt-allen!

britt-allen:
> Thanks @britt-allen!

Thank you for flagging and teaching me something new!

PR #135, opened 2024-03-26
Avoid row explosion in detector diagnostic tests

ian-r-rose:
This reworks the diagnostic set assignment so that the grain of diagnostics remains at the station_id level, avoiding row explosion. Previously, the join with the station metadata was matching every type of diagnostic, meaning that we got a ton of new rows for each station, making subsequent joins and aggregations much more costly. Now each diagnostic value is its own column.

Fixes #134 more permanently

### DIFF
--- a/transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
+++ b/transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
@@ -27,6 +27,43 @@ models:
         description: |
           The high-flow threshold for station diagnostics, above which the
           values are considered suspicious.
+      - name: high_occ_pct
+        description: |
+          The high occupancy percentage threshold for station diagnostics above which
+          the number of high occupancy values that occur at a station are suspect
+          No. of high occ samples > % of the max collected samples during the test period
+      - name: zero_occ_pct
+        description: |
+          The zero occupancy percentage threshold for station diagnostics above which
+          the number of zero occupancy values that occur at a station are suspect
+          No. of zero occ samples > % of the max collected samples during the test period
+      - name: flow_occ_pct
+        description: |
+          The percentage threshold for station diagnostics where the number of
+          flow = 0 and occupancy > 0 values that occur at a station are suspect
+          No. of flow-occ mismatch samples > % of the max collected samples during the
+          test period.
+      - name: occ_flow_pct
+        description: |
+          The percentage threshold for station diagnostics where the number of
+          occupany = 0 and flow > 0 values that occur at a station are suspect
+          No. of flow-occ mismatch samples > % of the max collected samples during the
+          test period.
+      - name: repeat_occ
+        description: |
+          The high-occupancy threshold for station diagnostics, above which the
+          values are considered suspicious
+          No. of repeated occupancy values > 5-min samples
+      - name: high_flow_pct
+        description: |
+          The percentage threshold for station diagnostics where the number of high flow
+          values that occur at a station are suspect
+          No. of high flow samples > % of the max collected samples during the test period
+      - name: zero_flow_pct
+        description: |
+          The percentage threshold for station diagnostics where the number of zero flow
+          values that occur at a station are suspect
+          No. of zero flow samples > % of the max collected samples during the test period
     tests:
       - dbt_utils.equal_rowcount:
           compare_model: ref('int_clearinghouse__most_recent_station_meta')


transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
```diff
@@ -24,30 +24,36 @@ det_diag_set_assign as (
...
 
     from {{ ref('int_clearinghouse__most_recent_station_meta') }}
 ),
 
-det_diag_threshold_values as (
+diagnostic_threshold_values as (
+    select *
+    from {{ ref('diagnostic_threshold_values') }}
+    pivot (avg(dt_value) for dt_name in ('high_occ', 'high_flow'))
+        as p (dt_set_id, dt_method, high_occupancy, high_flow)
```
ian-r-rose, line 40: Here's the important change, so that there is only one row per `dt_set_id`-`dt_method` pair.

transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
```diff
@@ -7,22 +7,30 @@ models:
...
+        description: |
+          The high-occupancy threshold for station diagnostics, above which the
+          values are considered suspicious.
+      - name: high_flow
+        description: |
+          The high-flow threshold for station diagnostics, above which the
+          values are considered suspicious.
+    tests:
+      - dbt_utils.equal_rowcount:
+          compare_model: ref('int_clearinghouse__most_recent_station_meta')
```
ian-r-rose, line 32: Adding a test here to guard against row explosion in this join

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
kengodleskidot: @ian-r-rose There are multiple threshold values that we want to bring in for mainline and ramps, here are some examples:
For mainline:
image
For Ramps:
image
I believe this model is only bringing in high_occupancy and high_flow threshold values. Is it more beneficial to bring in all threshold values here? The other threshold values (constant occupancy and percentages) will be used in the logic to determine the detector error type which we will want to add to the int_pems__diagnostic_samples_per_station.sql model.

ian-r-rose: Yes, this is only pivoting those two. I can also try to bring in the others. Do you know if we'll want all of them?

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
kengodleskidot, line 1: Great work @ian-r-rose! We will likely want to modify the final CTE in this model to include the tests to determine the detector error type.

ian-r-rose:
Okay @kengodleskidot, I've included the other threshold values, ready for another look!

transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
kengodleskidot: Do we need to add the name/description for all of the added threshold values here?

- high_occ_pct
- zero_occ_pct
- flow_occ_pct
- occ_flow_pct
- repeat_occ
- high_flow_pct
- zero_flow_pct

kengodleskidot (approved):
Looks great @ian-r-rose. Let me know if you want me to do any QA/QC before merging or if it's ok to do afterwards.

transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
ian-r-rose: Good catch! Some of them I don't actually have a good description for. Can you help me out with what `occ_flow_pct`, `flow_occ_pct`, and `repeat_occ` mean?

kengodleskidot: File updated in Github and changes committed

ian-r-rose: Thanks!

britt-allen: that's a clever test!

transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
```diff
@@ -7,22 +7,67 @@ models:
...
+      - name: station_id
         description: |
           An integer value that uniquely indenties the Station Metadata.
           Use this value to 'join' other clearinghouse files that contain Station Metadata.
       - name: district
         description: The district in which the VDS is located. Values are 1-12.
       - name: type
         description: Two character string identify the VDS type.
-      - name: det_diag_set_id
+      - name: station_diagnostic_set_id
```
britt-allen, line 18: appreciate the clarification, i'll be mindful of this going forward!

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
```diff
@@ -24,30 +24,57 @@ det_diag_set_assign as (
...
+        end as station_diagnostic_method_id
 
     from {{ ref('int_clearinghouse__most_recent_station_meta') }}
 ),
 
-det_diag_threshold_values as (
+diagnostic_threshold_values as (
+    select *
+    from {{ ref('diagnostic_threshold_values') }}
+    pivot (avg(dt_value) for dt_name in (
```
britt-allen, line 39: Why did you use the `avg()` function for `dt_value` as opposed to using the `any_value()` function?

britt-allen (commented):
Apologies for the late review, I know this is merged already but I still wanted to check out your work! Also thank you for doing this, next time I'd love more time to learn and try to do something like this myself before you submit the code especially if it's not urgently blocking project work. I think starting with a clear written issue description like you did  and then us discussing a timeline would be helpful for us both. Then if the timeline is not going to be met that can be an opportunity for you to jump in. What do you think?

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
britt-allen: I am not sure it notifies you because this is merged so tagging @ian-r-rose, but lmk?

ian-r-rose: Good question. I don't remember, I think I mostly just picked it randomly. I only expect one value for each of these (otherwise there is a problem with the underlying seed table). So `avg`, `any_value`, `mean`, etc should all give the same answer.

ian-r-rose:
> Apologies for the late review, I know this is merged already but I still wanted to check out your work! Also thank you for doing this, next time I'd love more time to learn and try to do something like this myself before you submit the code especially if it's not urgently blocking project work. I think starting with a clear written issue description like you did and then us discussing a timeline would be helpful for us both. Then if the timeline is not going to be met that can be an opportunity for you to jump in. What do you think?

That seems like a good plan to me. In this particular case, you were OOO for a few days, and this was blocking some of the downstream work on detector diagnostics, so I was motivated to resolve it quickly.

britt-allen:
Yes you're right I forgot I was OOO, thank you again for resolving this on my behalf and helping further the project work. Also I am not upset about this just in case it comes off that way. :)

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
britt-allen: I figured that was the case, in that any function would give the same value, but I wonder if avg is more expensive than any_value since it's performing actual arithmetic. I shall experiment on my own.

britt-allen: If I notice a significant performance diff I'll open a pr

ian-r-rose: You're probably right that average is slightly more expensive, I'm curious if you find anything. For large tables, I think you would be right that `any_value` has better performance. For this table, I suspect it would be really difficult to measure since it is only a few dozen rows long.

britt-allen: So `any_value()` is not supported with a pivot based on this thread [here](https://community.snowflake.com/s/question/0D50Z00008uVHTYSA4/anyvalue-does-not-work-on-pivot). Looks like `avg()` it is :)

ian-r-rose: Ah, that must have been why :)

PR #145, opened 2024-03-28
vmt sync 1

ZhenyuZhu-Caltrans:
vmt sync draft 1


ZhenyuZhu-Caltrans:
Ian, can you help review the VMT calculation sql in the performance_metrics folder before I build it for the first time. for me to avoid unexpected expense on snowflake. 
Thanks

ian-r-rose:
> Ian, can you help review the VMT calculation sql in the performance_metrics folder before I build it for the first time. for me to avoid unexpected expense on snowflake. Thanks

Hi @ZhenyuZhu-Caltrans , I'll take a look tomorrow! In the meantime, I encourage you to run the dbt models during your development process, and don't worry too much about running up the bill just yet (especially if you start with an XS warehouse). Running/testing frequently is an important part of the process! If it's too slow for your development workflow, you might try temporarily filtering the data to the last few days/weeks.

ZhenyuZhu-Caltrans:
Thanks a lot, Ian. 

I am running the code. It took 10 minutes and still going...

I will try to terminate it after 1 hour. 

Zhenyu

ZhenyuZhu-Caltrans:
After 15min with only 1.5% progress.
I called it off.
image

transform/models/intermediate/performance_metrics/_performance_metrics.yml
```diff
@@ -0,0 +1,136 @@
+# version: 2
```
ian-r-rose, line 1: Do you intend to have this fully commented out?

transform/models/intermediate/performance_metrics/int_performance_metrics__five_minute_vmt.sql
```diff
@@ -0,0 +1,54 @@
...
+        fivemin_agg.lane,
+        fivemin_agg.sample_date,
+        fivemin_agg.sample_timestamp,
+        -- VMT calculation (5 min agg volume * station coverage length)
+        fivemin_agg.volume * station_meta.length as vmt
+    from fivemin_agg
+    left join station_meta
+        on
+            fivemin_agg.id = station_meta.id
+            and fivemin_agg.lane = station_meta.lanes
```
ian-r-rose, line 50: I don't think this join is correct. This is joining the lane number on the total number of lanes. I don't think we actually need to join on lane here at all, grouping by `fivemin_agg.lane` should be enough.

ian-r-rose (commented):
Thanks @ZhenyuZhu-Caltrans, and sorry for the slow review! On the long execution times: I think you were running into the 5-minute aggregate model taking a long time to compute (which is very true!). Your VMT calculation should be significantly faster once that upstream model exists.

Next week we have a working session scheduled on how to more easily work with some of these larger tables. dbt has some tools to make this easier that we haven't done a great job communicating yet (namely cloning and deferral). But with some of those strategies in place, we should be able to get this workflow happening much more smothly

transform/models/intermediate/performance_metrics/_performance_metrics.yml
ZhenyuZhu-Caltrans: I have not yet worked on this file. 
I plan to work on this yml file once I debug the code.

transform/models/intermediate/performance_metrics/int_performance_metrics__five_minute_vmt.sql
ZhenyuZhu-Caltrans: @ian-r-rose thanks for the comment, I will look into this.

kengodleskidot:
I have added this metric to my other performance metrics since I need it to calculate Q and TTI.

PR #146, opened 2024-03-30
Detector status

kengodleskidot:
Detector status table which contains count information and a first attempt at determining a detector's status based on diagnostic threshold criteria. Needs additional QA/QC testing.


transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
```diff
@@ -0,0 +1,60 @@
+{{ config(
+    materialized="incremental",
+    cluster_by=['sample_date'],
+    unique_key=['station_id', 'sample_date', 'lane'],
+    snowflake_warehouse="transforming_xl_dev"
+) }}
```
ian-r-rose, line 6: At this stage, the `int_pem__diagnostic_samples_per_station` table is aggregated to the day, and, while moderately large, is no longer a beast (its <10 GB). I think that we can probably dispense with incremental models at this point. I checked out this branch and ran this model materialized as a table, and it took about two minutes on an XS warehouse

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
```diff
@@ -0,0 +1,60 @@
...
+        ddsc.*,
+        case
+            when ddsc.sample_ct = 0 or ddsc.sample_ct is null then 'Down/No Data'
+            -- # of samples < 60% of the max collected samples during the test period
+            -- max value: 2 samples per minute times 60 mins/hr times 17 hours in a day which == 1224
+            -- btwn 1 and 1224 is too few samples
+            when ddsc.sample_ct between 1 and (0.6 * (2 * 60 * 17)) then 'Insufficient Data'
+            when
+                set_assgnmt.station_diagnostic_method_id = 'ramp'
+                and ddsc.zero_vol_ct / (2 * 60 * 17) > (set_assgnmt.zero_flow_percent / 100) then 'Card Off'
```
ian-r-rose, line 35: Flagging for future discussion. This magic number of 2*60*17 shows up often enough that we might want to make it a project-wide variable. Doing that would make it easier to change in the future (if, for instance, we want to start counting for the full 24 hours)

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -2,23 +2,23 @@
     materialized="incremental",
     cluster_by=['sample_date'],
     unique_key=['station_id', 'sample_date', 'lane'],
-    snowflake_warehouse=get_snowflake_refresh_warehouse(small="XL")
+    snowflake_warehouse="transforming_xl_dev"
```
ian-r-rose, line 5: Let's talk about deferral strategies in our working sessions this week to ease some of the pain you are feeling for this large model @kengodleskidot!

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -65,22 +65,11 @@ samples_per_station as (
         COUNT_IF(source.volume > set_assgnmt.high_flow) as high_volume_ct,
         COUNT_IF(source.occupancy > set_assgnmt.high_occupancy) as high_occupancy_ct
 
-    from source
-    left join {{ ref('int_pems__det_diag_set_assignment') }} as set_assgnmt
+    from {{ ref('int_pems__det_diag_set_assignment') }} as set_assgnmt
+    left join source
```
ian-r-rose, line 69: Is there a reason you've swapped these? I think that the original ordering makes more sense

ian-r-rose (commented):
This is looking like it's in good shape @kengodleskidot! I have a few comments, and I think we can do some things to make your development experience a little smoother re: incremental models, but I'm liking where this is going.

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
```diff
@@ -55,6 +55,10 @@ detector_status as (
     from {{ ref('int_pems__det_diag_set_assignment') }} as set_assgnmt
     inner join det_diag_sample_count as ddsc
         on ddsc.station_id = set_assgnmt.station_id
+    --Look only at the previous day to determine a detectors status
+    --for development we use 2 days but the final code should be updated
+    --from -2 to -1 or use PREVIOUS_DAY in Snowflake
+    where ddsc.sample_date = DATEADD('day', -2, CURRENT_DATE())
```
ian-r-rose, line 61: `det_diag_sample_count` already filters by date. Did you see my comment that it may not be necessary to filter/incrementalized model at all?
?

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
kengodleskidot: Agreed

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
kengodleskidot: We want to include all active stations for the previous in the table but since the source doesn't have data for all stations we were not getting the full dataset. With this join type we are getting information for all stations. The continuing issue here is that we are still missing lane information for stations that are sending no data. We need the detector to station to controller relationship to complete this analysis correctly.

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
kengodleskidot: I was thinking that one of the where statements in the model could be redundant I just wasn't sure which one to keep.

ian-r-rose: I think we don't need to keep either, at this point. If the model becomes too painful (i.e., slow), we can restore it

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
```diff
@@ -1,5 +1,15 @@
 with
 
+detector_source as (
+    select
+        station_id,
+        detector_id,
+        detector_name,
+        detector_status,
+        length(lane_number) as lane_num
```
ian-r-rose, line 9: This is good, I had using `log10` in my brain, which is a worse approach

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
```diff
@@ -0,0 +1,56 @@
+{{ config(
+    materialized="incremental",
+    cluster_by=['sample_count_date'],
+    unique_key=['station_id', 'sample_count_date', 'lane_num'],
+    snowflake_warehouse="transforming_xl_dev"
+) }}
```
ian-r-rose, line 6: I don't think this should be an incremental model (at least, not yet). It's small enough and fast enough to compute that we should start off with it as a table.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -1,31 +1,18 @@
-{{ config(
-    materialized="incremental",
-    cluster_by=['sample_date'],
-    unique_key=['station_id', 'sample_date', 'lane'],
-    snowflake_warehouse=get_snowflake_refresh_warehouse(small="XL")
-) }}
-
```
ian-r-rose, line 7: This model does need to be incremental, can you restore the logic here?

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -1,31 +1,18 @@
...
-        {% if is_incremental() %}
-            -- Look back two days to account for any late-arriving data
-            and sample_date > (
-                select DATEADD(day, -2, MAX(sample_date)) from {{ this }}
-            )
-        {% endif %}
-        {% if target.name == 'dev' %}
-            and sample_date >= DATEADD('day', -14, CURRENT_DATE())
-        {% endif %}
+        and sample_date = DATEADD('day', -2, CURRENT_DATE())
```
ian-r-rose, line 7: Here as well

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -65,22 +52,18 @@ samples_per_station as (
...
-        -- max value: 2 samples per minute times 60 mins/hr times 17 hours in a day which == 1224
-        -- btwn 1 and 1224 is too few samples
-        COALESCE(sample_ct between 1 and (0.6 * (2 * 60 * 17)), false) as too_few_samples
-
-    from samples_per_station
+    from {{ ref('int_pems__det_diag_set_assignment') }} as set_assgnmt
+    left join source
+        on
+            source.id = set_assgnmt.station_id
+            and source.lane = set_assgnmt.lane_num
```
ian-r-rose, line 59: Does the set assignment ever differ between lane numbers? I'm not sure why we need to include the detector ID and lane number at this stage, which makes the join more expensive

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
```diff
@@ -1,5 +1,15 @@
 with
 
+detector_source as (
```
ian-r-rose, line 3: Do we need to include detector IDs at this stage? I don't think it's used anywhere, and is adding more complexity to the model here.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
kengodleskidot: The assignment does not change between lanes, it is specific to the station only currently. By adding the detector ID early I can filter out the lanes in the raw data that are not actual lanes in the freeway. I was able to reduce the table calculation to produce ~47000 rows associated with actual detectors instead of ~126000 rows, most of which were not actual detectors.

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
kengodleskidot: See previous comments and let me know your thoughts. I appreciate the comprehensive feedback @ian-r-rose!

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
kengodleskidot: The issue I see here is that we would be using the most current stations to process 14 days of raw data. Since stations can change daily this would give us incorrect or missing data for any days that the station status changed over the 14-day period. If we want to process multiple days in this model we will need to make sure we only process the stations that were active on a given date. I have some thoughts on how to do this but want to discuss and make sure it makes sense to move in that direction.

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
ian-r-rose: The other way to approach this is to rebuild the whole table from scratch every day. Once we are aggregated to the daily level (as we are here), this becomes much more tractable. We'd be paying a bit more for compute, but the benefit would be simpler to reason about models.

Put another way, I don't think we should pay the complexity cost of incremental models unless we absolutely have to.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
ian-r-rose: Interesting idea, I hadn't considered this... to my mind it's simpler to have the status model be non-incremental and have this be the incremental one. I'm curious what you view the benefit is of this being a view.

transform/seeds/districts.csv
```diff
@@ -0,0 +1,13 @@
+"DISTRICT_ID","DISTRICT_NAME","STATE_ID","DISTRICT_LATITUDE","DISTRICT_LONGITUDE"
+1,"Northwest",6,40.7245,-124.177
```
ian-r-rose, line 2: Are these centroids? Or central field offices? I wonder if it would make more sense to include the actual shapes of the districts as geography objects

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
ian-r-rose: If I understand you correctly, you are worried that if we look back in time too much, we might get out-of-date station metadata (e.g., if there is a station decommissioned).  This is a good thing to worry about! And we are almost certainly getting it wrong here by using the `most_recent_station_metadata` file. Any time we go back further in time than the most recent one, we are running the risk of joining with the wrong metadata.

The way that I think we should approach this is to not use `most_recent_station_metadata`, but instead use `int_clearinghouse__station_metadata`, which has the `_valid_from` and `_valid_to` columns. If we get the set assignment for a given date, this  These allow us to always join on the correct metadata for a given date.

When we set it up that way, it becomes always safe to rebuild the table without fear of getting the wrong metadata. And then it can be an incremental model. Having an approach that only looks two days back in the past means that if we ever need to rebuild older dates (if, e.g., our logic changes), we won't be able to.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
ian-r-rose: Doesn't `STATION_META` already contain a field for the number of lanes? That could be a way to reduce the number of rows produced without needing to join in the detector IDs.

transform/seeds/districts.csv
kengodleskidot: I am not sure but we do have geographical shapes associated with the districts located here: https://gis.data.ca.gov/datasets/0144574f750f4ccc88749004aca6eb0c/explore and https://gisdata-caltrans.opendata.arcgis.com/datasets/0144574f750f4ccc88749004aca6eb0c/explore?location=36.197987%2C-119.299050%2C6.31. If we can incorporate these instead of the point files that would be great since I am not sure the last time the district seed file was updated.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
kengodleskidot: @ian-r-rose The model does contain number of lanes per station, but some stations do not have detectors across all lanes. For example, we could have a 5-lane freeway with only 3 active detectors associated with the station. If we do not join on detectors we would be analyzing/reporting on lanes with no associated detector which is incorrect. It is rare but the difference is a few hundred detectors statewide.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
kengodleskidot: I believe we are on the same page with this topic. I believe it is safer to build our models using int_clearinghouse__station_metadata and incorporating the _valid_from and _valid_to columns. Let me take a shot at updating the code in this model, I should have something by COB 4/9/2024.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
kengodleskidot: Based on the other comments in this PR I prefer the idea of keeping the samples_per_station model incremental to account for late/missing incoming data.

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
kengodleskidot: We will want to keep a history of the status value associated with each active station on a daily basis. The current table in the database has 284099697 rows and will only grow. The historical status information is useful for Caltrans staff because we use this as a metric to decide which detectors to assign resources to for troubleshooting or replacement. There are other uses as well. If we do not update this table using the incremental model is there another way to accomplish this?

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
ian-r-rose: Ah, right, you mentioned this but I forgot!

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
britt-allen: this is helpful discussion, glad this is occurring hear rather than orally, thank you both!

britt-allen:
@kengodleskidot can you be sure to resolve conversations that are resolved, please?

transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
```diff
@@ -110,9 +135,15 @@ models:
         description: |
           Counts the number of occupancy values in a lane that exceeded the high value threshold
           detector diagnostic parameter for a station.
-      - name: too_few_samples
```
britt-allen, line 113: where do we preserve or point to the logic that is used in calculating these statuses? i think that needs to live in documentation somewhere, if not the data dictionary, then this could be a good case for adding to your docs here: https://cagov.github.io/caldata-mdsa-caltrans-pems/

britt-allen (commented):
I know this is still a WIP but this is looking great so far. Thanks for all of your work on this Ken.

transform/seeds/districts.csv
ian-r-rose: We can definitely incorporate the actual shapes if it would be helpful in some of the downstream analyses. For the time being, I think we can move forward with this seed, then update it later, how does that sound?

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
ian-r-rose: I agree that we want to keep the history of the status, I'm suggesting that we re-calculate the entire history every day. If we have ~20k detectors running for ~10K days, this works out to a ~200 million row table (I think that's the same order of magnitude that you have above).

In the broad scheme of things it might be okay to re-compute a 200mm row table as part of our normal pipeline. I'm suggesting that the extra compute price here might be worth the simplicity of not having this be an incremental model.

transform/seeds/districts.csv
kengodleskidot: Sounds great!

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
kengodleskidot: Detector id's are left off until we identify a consistent data source

kengodleskidot:
FYI dbt Cloud error:
image

transform/dbt_project.yml
britt-allen, line 1: Who wrote this beautiful YAML? 🤠

britt-allen (approved):
Great work @kengodleskidot!

ian-r-rose:
This will not pass in prod because there is a dev warehouse in the config. Can you make a quick fix @britt-allen?

britt-allen:
> This will not pass in prod because there is a dev warehouse in the config. Can you make a quick fix @britt-allen?

Omg, yes

PR #148, opened 2024-04-09
Constant occupancy britt

### LINKED ISSUE #83: PeMS detector health diagnostics: occupancy is constant
Implement Test number 6 from the [PeMS User Guide](https://pems.dot.ca.gov/Papers/PeMS_Intro_User_Guide_v6.pdf):

image

**Data sources**: 30 second raw data

britt-allen:
fixes #83


transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -65,22 +52,48 @@ samples_per_station as (
...
+    from source
 )
 
-select * from det_diag_too_few_samples
+select
+    sps.*,
+    dfc.district_feed_working,
+    ois.sample_timestamp,
+    ois.occupancy,
+    COALESCE((SUM(case when ois.occupancy = ois.prev_occ then 1 else 0 end) / COUNT(*)) >= 0.6, false) as constant_occupancy
```
kengodleskidot, line 93: Is this where are checking to see if we have a sufficient number of samples to make a constant occupancy decision?

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -65,22 +52,48 @@ samples_per_station as (
...
+from samples_per_station as sps
+inner join district_feed_check as dfc
```
kengodleskidot, line 95: FYI based on the current method of aggregating the raw data samples we are no longer doing the district feed check CTE here.

kengodleskidot (commented):
Good progress and FYI there have been some recent changes in the latest detector status models based on our conversation with Ian yesterday that I noted. Let's go over this in more detail in our meeting on 4/12 😊

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
britt-allen: yes

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -65,22 +52,48 @@ samples_per_station as (
...
+    group by samples_per_station.district
+),
+
+occupancy_is_constant as (
+    select
+        source.id,
+        source.sample_timestamp,
+        source.occupancy,
+        LAG(source.occupancy) over (
+            partition by FLOOR((DATE_PART('epoch', source.sample_timestamp) - DATE_PART('epoch', DATEADD(second, 14400, source.sample_timestamp))) / 14400) order by source.sample_timestamp
```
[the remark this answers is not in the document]
britt-allen: ty for the reminder!

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
```diff
@@ -78,6 +80,31 @@ samples_per_station as (
...
+        f.sample_date,
+        f.occupancy,
+        LAG(f.occupancy) over (order by f.sample_date) as previous_occupancy
+    from {{ ref("int_clearinghouse__five_minute_station_agg") }} as f
+),
+
+constant_occupany_check as (
+    select
+        *,
+        COUNT(occupancy = previous_occupancy)
```
ian-r-rose, line 97: I don't think this is the right condition. `COUNT` counts the non-null values. I think you need `COUNT_IF`.

I also don't think this logic does quite what we want with `COUNT_IF`: if there are several regions with repeat occupancies, but they are separated by regions without repeat occupancies, this could still return true. For instance, if we have ten periods of three repeats, this would result in a repeat occupancy diagnosis, even though we never got 28 in a row.

What about the logic we discussed where we sum the differences of the absolute values?

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
ian-r-rose, line 1: Since this is an incremental model, I don't think we want to bring anything to do with the 5 minute station agg here. As you have it, it's doing the incremental logic for the 30 second data, but no incremental logic for the 5-minute data.

It's probably best to have this repeat-occ calculation in a separate model

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
britt-allen: I struggled with that logic so I tried something else that was more intuitive to me. I'll spend some more time thinking about it. Let me know if you're available for a working session on Tuesday.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,51 @@
...
+
+calculate_occupancy_difference as (
+    select
+        id,
+        sample_timestamp,
+        sample_date,
+        occupancy,
+        case
+            when occupancy != 0
+                then occupancy - LAG(occupancy) over (partition by id, sample_timestamp order by sample_timestamp)
```
ian-r-rose, line 37: If we're partitioning by `sample_timestamp`, then `sample_timestamp` will be the same within each value (same as `group by`). So then it doesn't make sense to order by `sample_timestamp`.

I think you want to partition by ID and date (or possibly just ID), and perform the lag over that.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,51 @@
...
+        end as occupancy_difference
+    from source
+    group by id, sample_timestamp, sample_date, occupancy
+    order by sample_timestamp
```
ian-r-rose, line 41: I don't think this group by makes sense for this calculation. There are no aggregation functions above, so it's a do nothing group by.

transform/models/intermediate/detector_health_diag/int_pems__diagnostic_samples_per_station.sql
ian-r-rose, line 1: Sorry if I was unclear before: I don't think we could include the constant occupancy in this model at all. It's a different enough computation that I don't think it's worth the price of joining here. I think it's better to keep this model as simple as we can, and do any joining logic in the status table.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,51 @@
...
+
+)
+
+select
+    *,
+    ABS(occupancy_difference) as constant_occupancy
+from calculate_occupancy_difference
+qualify
+    SUM(constant_occupancy) over (partition by id, sample_timestamp order by sample_timestamp rows between 47 preceding and current row) = 0
```
ian-r-rose, line 50: Same here

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,51 @@
...
+),
+
+calculate_occupancy_difference as (
+    select
+        id,
+        sample_timestamp,
+        sample_date,
+        occupancy,
+        case
+            when occupancy != 0
```
ian-r-rose, line 36: I *think* this check makes sense to me (we are okay with several hours of constant occupancy if the occupancy is zero). But I'm not sure if it has the effect you are looking for. When occupancy *is* zero, then `occupancy_difference` is null (because there is no other branch of the `case` statement). Is that what we want? 

In general, I'm wary of `case` statements that could be more simply computed with arithmetic operations, which should be cheaper to evaluate.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,51 @@
...
+    group by id, sample_timestamp, sample_date, occupancy
+    order by sample_timestamp
+
+)
+
+select
+    *,
+    ABS(occupancy_difference) as constant_occupancy
+from calculate_occupancy_difference
+qualify
```
ian-r-rose, line 49: Ooh, `qualify` spotted in the wild! So, the intention here is to only keep observations of constant occupancy, and throw away anything else?

ian-r-rose (commented):
Looking close!

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
britt-allen: That was not the intention. I mostly wanted to practice qualify after learning it earlier, i see it was a misuse now. Thank you!

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
britt-allen: I don't think the output is what I wanted so I updated the code. Thank you!

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
britt-allen: oopsie, for some reason i thought of the window function as an aggregrate function. i realize now that i can aggregate windowed data, but that's not what i was doing so you're right that the grouping doesnt make sense. fixed.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
britt-allen: i partitioned by id and ordered by date since window functions need the order by clause. ty!

britt-allen:
not understanding the latest dbt error`Expression type does not match column data type, expecting FLOAT but got BOOLEAN for column CONSTANT_OCCUPANCY`, when i run the exact code in snowflake it produces results, there's no error

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
ian-r-rose: Yep, exactly right. Windows are often similar to group by statements, except windows can have logic about the ordering within a partition

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
ian-r-rose: I think it actually *could* be the case that a qualify statement is useful here, but I'm not certain. At the end of this, we want a table where there is a single row for each lane-station-date combination, and a column that indicates whether we got a constant occupancy at any point during the day. But let's maybe get to that stage first and consider whether a qualify statement could help simplify the logic.

ian-r-rose:
> not understanding the latest dbt error`Expression type does not match column data type, expecting FLOAT but got BOOLEAN for column CONSTANT_OCCUPANCY`, when i run the exact code in snowflake it produces results, there's no error

I'll take a look in the morning!

transform/models/intermediate/detector_health_diag/int_pems__detector_status.sql
```diff
@@ -59,6 +64,21 @@ detector_status as (
...
+    left join {{ ref('int_pems__constant_occupancy') }} as co
+        on
+            sps.station_id = co.id and sps.lane = co.lane and sps.sample_date = co.sample_date
+            {% if is_incremental() %}
+            -- Look back two days to account for any late-arriving data
+            and sps.sample_date > (
+                select
+                    DATEADD(day, {{ var("incremental_model_look_back") }}, MAX(sps.sample_date))
+                from {{ this }}
+            )
```
ian-r-rose, line 76: This does something weird where you use incremental logic in `int_pems__constant_occupancy`, but select the _full_ table from `int_pems__diagnostic_samples_per_station`, rather than doing any incrementality there. So even though you have tagged this as an incremental model here (cf #161), it doesn't actually implement incrementality for the main status table.

I also find it weird to have this filtering clause in the join condition. I suggest moving it into a top-level CTE filtering `int_pems__constant_occupancy` rather than here.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,70 @@
...
+),
+
+sum_occupancy_delta as (
+    select
+        *,
+        ABS(occupancy_delta) as abs_val_occupancy_delta,
+        SUM(abs_val_occupancy_delta)
+            over (
+                partition by id, lane, sample_date
+                order by sample_timestamp rows between 47 preceding and current row
```
ian-r-rose, line 55: Let's add a comment indicating where this `47` comes from.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,70 @@
...
+        sample_timestamp,
+        sample_date,
+        lane,
+        occupancy,
+        occupancy
+        - LAG(occupancy)
+            over (partition by id, lane, sample_date order by sample_timestamp)
+            as occupancy_delta
+    from source
+    order by sample_timestamp
```
ian-r-rose, line 45: This `order by` is unnecessary, and may cause extra work.

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,70 @@
...
+        *,
+        ABS(occupancy_delta) as abs_val_occupancy_delta,
+        SUM(abs_val_occupancy_delta)
+            over (
+                partition by id, lane, sample_date
+                order by sample_timestamp rows between 47 preceding and current row
+            )
+            as abs_val_occupancy_delta_summed
+    from calculate_occupancy_delta
+    order by sample_timestamp
```
ian-r-rose, line 59: Same here

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
ian-r-rose, line 1: Can you add some yaml for this model?

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
```diff
@@ -0,0 +1,70 @@
...
+)
+
+select
+    id,
+    sample_date,
+    lane,
+    MIN(abs_val_occupancy_delta_summed) as min_occupancy_delta
+from sum_occupancy_delta
+where occupancy != 0 or occupancy is not null
```
ian-r-rose, line 68: I think this `where` clause could actually be a good candidate for using `qualify` in the previous CTE. Basically, we want to filter the results of the 47-row look back to remove ones where occupancy is 0 or `null`. Doing that *might* improve the performance of this group by (or it might not)

ian-r-rose (approved):
Thanks @britt-allen, this looks good to me. I think the only outstanding thing is to add some YAML docs for the new model

transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
```diff
@@ -131,6 +131,10 @@ models:
         description: |
           Counts the number of occupancy values in a lane that exceeded the high value threshold
           detector diagnostic parameter for a station.
+      - name: constant_occupancy
+        description: |
+          Counts the number of samples of occupancy for a 4-hour window that are constant and
```
kengodleskidot, line 136: It would be beneficial to mention that the occupancy values being evaluated should be non-zero.

kengodleskidot (approved):
Great job @britt-allen! I had a minor comment in the yml for the constant occupancy definition.

transform/models/intermediate/detector_health_diag/_detector_health_diag.yml
britt-allen: Roger that!

transform/models/intermediate/detector_health_diag/int_pems__constant_occupancy.sql
britt-allen: I forgot! Thanks for the second reminder below. It's there now.

PR #158, opened 2024-04-12
Revert "Detector status"

britt-allen:
Reverts cagov/caldata-mdsa-caltrans-pems#146


ian-r-rose:
I don't think we need to revert the PR, just fix up the warehouse in the config

britt-allen:
This was unnecessary

PR #159, opened 2024-04-12
fixed the warehouse logic in the config block and variable quote issues

britt-allen:


kengodleskidot (approved):
Looks great, thank you @britt-allen!

Issue #163, opened 2024-04-16
Use `big` and `small` parameters in `get_snowflake_refresh_warehouse()`


ian-r-rose:
Over in the `get_snowflake_refresh_warehouse()`, I added parameters for "big" and "small" to customize what warehouse sizes to use when running incremental models incrementally or when running them with a full refresh.

But I forgot to actually use those parameters! We need to add them here:

https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/eb4b661ad777feb5820bae2b925da1a0bb4882d2/transform/macros/get_snowflake_refresh_warehouse.sql#L8-L12

kengodleskidot:
Closed with PR #197

PR #164, opened 2024-04-17
Performance measures ken g

kengodleskidot:
Created a new schema, model and associated YML file to calculate the VMT, VHT, Q, TTI performance metrics. Also updated the five-minute aggregate model to count the number of samples collected for future use in data fidelity metrics.

### DIFF
--- a/transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
+++ b/transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
@@ -32,8 +32,16 @@ with station_raw as (
                     )
                 from {{ this }}
             )
+            {% if target.name != 'prd' %}
+                and sample_date
+                >= dateadd(
+                    day,
+                    {{ var("dev_model_look_back") }},
+                    current_date()
+                )
+            {% endif %}
     {% elif target.name != 'prd' %}
-        where sample_date >= dateadd('day', {{ var("dev_model_look_back") }}, current_date())
+        where sample_date >= dateadd(day, {{ var("dev_model_look_back") }}, current_date())
     {% endif %}
 ),
 
@@ -62,7 +70,8 @@ aggregated_speed as (
         --a conversion to get miles per hour (5280 ft / mile and 12
         --5-minute intervals in an hour).
         --The following code may be used if we want to use speed from raw data
-        --coalesce(speed_raw, ((volume * 22) / nullifzero(occupancy) * (1 / 5280) * 12))
+        --coalesce(speed_raw, ((volume * 22) / nullifzero(occupancy) 
+        --* (1 / 5280) * 12))
         --    as speed
         (volume * 22) / nullifzero(occupancy) * (1 / 5280) * 12 as speed
     from aggregated
--- a/transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
+++ b/transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
@@ -1,8 +1,8 @@
 {{ config(
     materialized="incremental",
     cluster_by=["sample_date"],
-    unique_key=["ID", "LANE", "SAMPLE_TIMESTAMP"],
-    snowflake_warehouse=get_snowflake_refresh_warehouse()
+    unique_key=["id", "lane", "sample_timestamp"],
+    snowflake_warehouse = get_snowflake_refresh_warehouse(small="XL")
 ) }}
 
 with station_raw as (
@@ -25,13 +25,13 @@ with station_raw as (
         where
             sample_date > (
                 select
-                    dateadd(day, {{ var("incremental_model_look_back") }}, max(sample_date))
+                    dateadd(
+                        day,
+                        {{ var("incremental_model_look_back") }},
+                        max(sample_date)
+                    )
                 from {{ this }}
             )
-            {% if target.name != 'prd' %}
-                and sample_date
-                >= dateadd('day', {{ var("dev_model_look_back") }}, current_date())
-            {% endif %}
     {% elif target.name != 'prd' %}
         where sample_date >= dateadd('day', {{ var("dev_model_look_back") }}, current_date())
     {% endif %}
@@ -43,8 +43,12 @@ aggregated as (
         sample_date,
         sample_timestamp_trunc as sample_timestamp,
         lane,
+        count_if(volume is not null and occupancy is not null)
+            as sample_ct, --Number of raw data samples
         sum(volume) as volume, -- Sum of all the flow values
         avg(occupancy) as occupancy -- Average of all the occupancy values
+        -- avg(speed) as speed_raw -- This code could be used if we use 
+        -- actual speeds reported from raw data 
     from station_raw
     group by id, lane, sample_date, sample_timestamp_trunc
 ),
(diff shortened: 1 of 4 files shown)


transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -1,8 +1,8 @@
 {{ config(
     materialized="incremental",
     cluster_by=["sample_date"],
-    unique_key=["ID", "LANE", "SAMPLE_TIMESTAMP"],
-    snowflake_warehouse=get_snowflake_refresh_warehouse()
+    unique_key=["id", "lane", "sample_timestamp"],
+    snowflake_warehouse = get_snowflake_refresh_warehouse(small="XL")
```
ian-r-rose, line 5: This "small" parameter won't work until #163 is fixed

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -25,13 +25,13 @@ with station_raw as (
...
+                        day,
+                        {{ var("incremental_model_look_back") }},
+                        max(sample_date)
+                    )
                 from {{ this }}
             )
-            {% if target.name != 'prd' %}
-                and sample_date
-                >= dateadd('day', {{ var("dev_model_look_back") }}, current_date())
-            {% endif %}
```
ian-r-rose, line 34: I think we do want this check, for the dev experience (we don't want to build the full model in dev).

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -0,0 +1,70 @@
...
+        where
+            sample_date > (
+                select
+                    dateadd(
+                        day,
+                        {{ var("incremental_model_look_back") }},
+                        max(sample_date)
+                    )
+                from {{ this }}
+            )
```
ian-r-rose, line 23: As above, it might be helpful to have a `dev_model_look_back` clause for when `is_incremental()` is true

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -0,0 +1,70 @@
...
+
+station_meta as (
+    select
+        fma.*,
+        sm.length,
+        sm.type,
+        sm._valid_from as station_valid_from,
+        sm._valid_to as station_valid_to
+    from {{ ref ('int_clearinghouse__station_meta') }} as sm
+    inner join five_minute_agg as fma
```
ian-r-rose, line 37: I usually think it's clearer to join the "fact" table with the "dimension" table, rather than the other way around. That way the table with the [grain](https://docs.getdbt.com/terms/grain) is always first in the join.

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -0,0 +1,70 @@
...
+                        max(sample_date)
+                    )
+                from {{ this }}
+            )
+    {% elif target.name != 'prd' %}
+        where sample_date >= dateadd('day', {{ var("dev_model_look_back") }}, current_date())
+    {% endif %}
+),
+
+station_meta as (
```
ian-r-rose, line 29: This isn't really a metadata table, it's a five minute agg table with metadata attached. I think we can make the CTE name a bit clearer

ian-r-rose (commented):
This is looking great @kengodleskidot!

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -0,0 +1,70 @@
...
+        vmt / nullifzero(vht) as q_value
+    from vmt_vht_metrics
+),
+
+tti_metric as (
+    select
+        *,
+        60 / nullifzero(q_value) as tti
+    from q_metric
+)
```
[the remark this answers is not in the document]
kengodleskidot: @ian-r-rose  since I need vmt and vht to calculate Q and Q to calculate TTI I thought these calculations would need to be done in a previous CTE in order to use the results. Can I perform the calculations all in the same CTE for these metrics since they all rely on one another?

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
kengodleskidot: Understood!

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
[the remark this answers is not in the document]
ian-r-rose: Good point! This is a case where different SQL dialects differ. I think that SQL Server and Oracle don't support reusing column aliases in that way, but some newer SQL dialects like Snowflake and DuckDB do. So while I do think it would work to combine them into one CTE, if you think it would be clearer and more portable to write it like this, I'm happy to keepo it this way.

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
kengodleskidot: Doesn't the code below (lines 35-37) perform the dev calculation we want or am I mistaken?

ian-r-rose: It does, but only when `is_incremental()` is false. This snippet does the same for when `is_incremental()` is true. There are probably other ways of setting this up so it's less duplicative, but they do handle different cases

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
kengodleskidot: Good catch! I'll add the code associated with this process now.

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
kengodleskidot: Updated

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
kengodleskidot: Thank you!

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
kengodleskidot: Got it. I added back the additional if statement in the is incremental block and left the code below as is.

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -0,0 +1,81 @@
...
+    inner join {{ ref ('int_clearinghouse__station_meta') }} as sm
+        on
+            sm.id = fma.id
+            and sm._valid_from <= fma.sample_date
+            and
+            (
+                sm._valid_to > fma.sample_date
+                or sm._valid_to is null
+            )
+),
```
ian-r-rose, line 55: As we are building more metrics on top of the five minute agg, this join will be used in multiple places. I wonder if it would be worth creating a new model for it. It's an unfortunately expensive join, but it's hard to see a way around having some more information about the detectors in these computations.

No action item here, this logic looks good to me, just thinking about the future state.

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -0,0 +1,81 @@
...
+--         vmt / nullifzero(vht) as q_value
+--     from vmt_vht_metrics
+-- ),
+
+-- tti_metric as (
+--     select
+--         *,
+--         60 / nullifzero(q_value) as tti
+--     from q_metric
+-- )
```
ian-r-rose, line 79: Let's delete this unused code

ian-r-rose (commented):
This is looking really good to me. I see one tiny formatting issue that the pre-commit check is flagging, but I don't have other concerns.

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
kengodleskidot: I was thinking about the creation of a separate model which contains station meta data for all active stations on a given day and (probably) making it incremental. The model would be run just after midnight to account for any station updates done the previous day and would be used for any calculations such as detector health and performance metrics. The key piece we are still missing in our models is including stations where raw data is not being reported. This will come into play even more when we work on imputation. An advantage of having the new model would be a single source of truth for active stations that day and making sure all associated calculations are done on the active stations no matter which model is being run. I have run into situations in the current PeMS data sets where different station counts are returned between detector health and performance metric data on the same date when it should be consistent.

ian-r-rose (approved):
Looks great, thanks @kengodleskidot!

britt-allen (commented):
This is great Ken, apologies for the late review!

Issue #181, opened 2024-04-22
Introduce `pems_clearinghouse_start_date` variable


ian-r-rose:
As we are in the thick of defining some of our base, large models, we are recomputing some quite large tables with some regularity. This is both expensive and time consuming.

What do we think about introducing a new variable `pems_clearinghouse_start_date` which can be set to, e.g., 2023-01-01 for the time being? We can filter our base models according to this date, and once the dust settles a bit, move it back to the actual start of the historical PeMS data.

kengodleskidot:
@ian-r-rose I agree with your approach and like the idea. The proposed start date should be fine for where we are at in the project and it gives us a full year of data which I believe is adequate for the time being to perform QA/QC. There may be other anomalies farther back in time but these will be easier to identify once we have some visualization and reporting tools.

ian-r-rose:
@kengodleskidot would you be willing to take this on?

PR #183, opened 2024-04-23
Active Station Model, Incremental Updates and District Feed Down Detector Status

kengodleskidot:
This pull request updates the detector_status model to be incremental and updates the samples_per_station model to use the latest incremental logic. A new active_stations model has been built which is currently a view that contains the most recent weeks list of active stations on a daily basis. This has been implemented in the det_diag_set_assignment model so stations without raw data are still included in the detector_status model. I am still determining why extra records are being created in the detector_status model but the rest of the QA/QC looks good when compared to the current PeMS data.


kengodleskidot:
> This pull request updates the detector_status model to be incremental and updates the samples_per_station model to use the latest incremental logic. A new active_stations model has been built which is currently a view that contains the most recent weeks list of active stations on a daily basis. This has been implemented in the det_diag_set_assignment model so stations without raw data are still included in the detector_status model. I am still determining why extra records are being created in the detector_status model but the rest of the QA/QC looks good when compared to the current PeMS data.

UPDATE - The extra records are due to 8 detector/lane values being reported for each station based on the raw data analysis in the samples_per_station model. The extra records should be addressed once we have the detectors per station data and the detector_status model join is updated to include detector/lane.

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
```diff
@@ -0,0 +1,27 @@
+with
+
+recursive date_range as (
+    select dateadd(day, {{ var("dev_model_look_back", current_date) }}, current_date) as active_date
```
britt-allen, line 4: Can you explain what `current_date` is actually doing inside the variable? It seems like incorrect variable syntax.

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
```diff
@@ -0,0 +1,27 @@
...
+
+station_meta as (
+    select * from {{ ref("int_clearinghouse__station_meta") }}
+),
+
+active_station as (
+    select
+        dr.*,
+        sm.*,
+        coalesce(sm._valid_to, current_date + 1) as valid_to
```
britt-allen, line 19: why are we producing a second `valid_to` column? a comment explaining this logic and decision may be helpful here. when would we use  `_valid_to` column versus the `valid_to` column? how are they different? do we need to retain both?

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
```diff
@@ -0,0 +1,27 @@
+with
+
+recursive date_range as (
```
britt-allen, line 3: Can you explain the intent behind this `recursive date_range` CTE and why we need an `active_date` column in addition to the `_valid_to` and `_valid_from` columns we already have?

britt-allen (commented):
Everything looks good @kengodleskidot I just have a few questions and 1 suggestion for the `int_clearinghouse__active_stations.sql` model.

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
ian-r-rose, line 4: I don't think we should have `dev_model_look_back` appear in any "production" code. That is to say, I think the only place we should see this variable is if it is wrapped in a `if target.name != 'prd'`.

In general, we should try to make it so that our final tables don't depend on the date on which it is run (incrementality aside, of course). Why not make this model go all the way back to the start of the computation (whether the full history, or clipped by #181 )?

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
```diff
@@ -31,7 +32,7 @@ station_diagnostic_set_assign as (
             else 'mainline'
         end as station_diagnostic_method_id
 
-    from {{ ref ('int_clearinghouse__station_meta') }}
+    from {{ ref ('int_clearinghouse__active_stations') }}
```
ian-r-rose, line 35: I may be misunderstanding, but it looks like this is a fairly major change to the model here: before it worked on the entire history, and with this change it only goes back a week.

In general, I recommend *against* doing anything that only goes back a few days. That makes these models very difficult to rebuild on-demand. If the models are large and taking too much time to rebuild, then we can do it explicitly with an incremental materialization. But this approach will typically make it quite difficult to do more historical analyses, or to make other tables depend on this one, as different tables will cover very different time periods.

If we want to move faster and with more agility, I think it may be better to take the approach of #181 and do some filtering right at the very base of the model hierarchy, then remove that filter once things have stabilized a bit.

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
kengodleskidot: Agreed and I am looking into the date_spine function mentioned in Ian's comment

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
kengodleskidot: Since _valid_to is null for currently active stations I needed to have an actual date value for the on/between statement. I only need the one variable so I can eliminate one of them. Thank you!

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
kengodleskidot: The recursive date range is used to establish dates for which we want to know the active stations. This is needed since the raw data is not included for all active stations in any day. In order to determine imputed values, performance measures, detector health, etc. for stations with no raw data this was the method I decided. I am open for alternatives to eliminate this CTE if possible.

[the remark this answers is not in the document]
kengodleskidot: Is this the macro you are referring to?
image

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
kengodleskidot: Agreed and thanks for the discussion! I will work on implementing this model with date_spine to address the date issue.

kengodleskidot:
> > UPDATE - The extra records are due to 8 detector/lane values being reported for each station based on the raw data analysis in the samples_per_station model. The extra records should be addressed once we have the detectors per station data and the detector_status model join is updated to include detector/lane.
> 
> I've gone back and forth on this a couple of times, but now I'm back to thinking that we _should_ join with `int_clearinghouse__station_meta` after aggregating to get rid of all the extra non-existent stations in, e.g., lane 8. It's an expensive join, so we'd want to be careful how we construct it so we do it as efficiently as possible, but seems unavoidable. What do you think @kengodleskidot?

I think there is a benefit to reducing the number of aggregated data points to at most the number of lanes associated with a station but the number of lanes does change with time. The active_stations model was intended to be the model joined against the aggregated values since it has station data where raw data does not exist but I think there is a solution here that can be used across multiple schema. Let's discuss in more detail as part of the imputation discussion on 4/30/24.

transform/models/intermediate/clearinghouse/int_clearinghouse__active_stations.sql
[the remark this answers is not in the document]
ian-r-rose: Yes, that's the one! I think it probably uses something like a recursive CTE internally, but would (in my opinion) simplify this query a bit.

transform/models/intermediate/detector_health_diag/int_pems__det_diag_set_assignment.sql
britt-allen: Are you still looking for clarity from Ian on this question? @kengodleskidot

britt-allen:
> > > UPDATE - The extra records are due to 8 detector/lane values being reported for each station based on the raw data analysis in the samples_per_station model. The extra records should be addressed once we have the detectors per station data and the detector_status model join is updated to include detector/lane.
> > 
> > 
> > I've gone back and forth on this a couple of times, but now I'm back to thinking that we _should_ join with `int_clearinghouse__station_meta` after aggregating to get rid of all the extra non-existent stations in, e.g., lane 8. It's an expensive join, so we'd want to be careful how we construct it so we do it as efficiently as possible, but seems unavoidable. What do you think @kengodleskidot?
> 
> I think there is a benefit to reducing the number of aggregated data points to at most the number of lanes associated with a station but the number of lanes does change with time. The active_stations model was intended to be the model joined against the aggregated values since it has station data where raw data does not exist but I think there is a solution here that can be used across multiple schema. Let's discuss in more detail as part of the imputation discussion on 4/30/24.

Was there a resolution reached on this?

britt-allen (approved):
This all looks good to me, thank you Ken! I have a few outstanding questions as it appears from the discussion there may or may not be some outstanding decisions to be made and implemented. @kengodleskidot

kengodleskidot:
> This all looks good to me, thank you Ken! I have a few outstanding questions as it appears from the discussion there may or may not be some outstanding decisions to be made and implemented. @kengodleskidot

@britt-allen I believe there is still a quick discussion to have about the use of active_stations instead of int_clearinghouse__station_meta in the set_assignment model and it's effect on the detector_status model. The current PR captures all stations that are considered active in a day (19540 for today 5/3/24) using the active_stations model in detector_status whereas joining on the int_clearinghouse__station_meta in detector_status leaves out active stations that have no raw data being sent from the district Traffic Managment Center (TMC) feeds.

Issue #165, opened 2024-04-17
Speed Calculation Update Discussion


kengodleskidot:
As we start to work on performance metrics and eventually move into reports/visualizations I would like to explore updating the speed calculation to minimize outliers in our data based on the simplified speed calculation we are currently using. I propose using the following update to the logic to determine speed:

If speed is being reported by a detector use the reported speed
If no speed is being reported then calculate speed using the simplified speed equation
     If a speed calculation using the simplified speed equation is over 90mph or null use a free flow speed of 65mph

I believe this change will give us speed and performance metrics more in line of what we would expect to see in the real world. I am open to the group's thoughts and inputs on this topic since it would involve recalculating the five-minute aggregate table.

ian-r-rose:
This sounds like a reasonable plan @kengodleskidot. Based on your logic, I think we may want to make the 5-minute aggregate table compute *just* the average speed based on that reported by the detector. That would help keep this expensive-to-calculate model as simple and stable as possible. Then your follow-up logic could be done in downstream models. Does this work for you?

>  If a speed calculation using the simplified speed equation is over 90mph or null use a free flow speed of 65mph

I would probably prefer to have it as null rather than a made up number.

ian-r-rose:
> @ian-r-rose since we are only getting speed values from a small group of detectors what model would be best to perform the simplified speed calculation if not the 5-min aggregate model?

We could put some of the more complex speed reasoning in your 5-minute performance metrics model. It's not the only choice, but shouldn't be too hard to add there

kengodleskidot:
FYI @mmmiah is working on updating the associated speed models

britt-allen:
> This sounds like a reasonable plan @kengodleskidot. Based on your logic, I think we may want to make the 5-minute aggregate table compute _just_ the average speed based on that reported by the detector. That would help keep this expensive-to-calculate model as simple and stable as possible. Then your follow-up logic could be done in downstream models. Does this work for you?
> 
> > If a speed calculation using the simplified speed equation is over 90mph or null use a free flow speed of 65mph
> 
> I would probably prefer to have it as null rather than a made up number.

I agree with Ken and agree with Ian's edit regarding preserving nulls rather than making up a number.

PR #188, opened 2024-04-25
Speed update april 2024

mmmiah:
This is the updated five minutes weighted speed calculation


transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -53,28 +54,27 @@ aggregated as (
...
-        --The following code may be used if we want to use speed from raw data
-        --coalesce(speed_raw, ((volume * 22) / nullifzero(occupancy)
-        --* (1 / 5280) * 12))
-        --    as speed
-        (volume * 22) / nullifzero(occupancy) * (1 / 5280) * 12 as speed
+        case
+            when
+                --average occupancy should not be null and zero
+                (average_occupancy is not null and average_occupancy != 0)
+                and (volume_sum is not null and volume_sum != 0) --sum of the volume should not be null and zero
```
britt-allen, line 73:
```suggestion
-- sum of the volume should not be null and zero
                and (volume_sum is not null and volume_sum != 0) 
```
we like to consistently place comments _before_ the logic they are about so people can read the context before the logic.

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -53,28 +54,27 @@ aggregated as (
...
-        --a conversion to get miles per hour (5280 ft / mile and 12
-        --5-minute intervals in an hour).
-        --The following code may be used if we want to use speed from raw data
-        --coalesce(speed_raw, ((volume * 22) / nullifzero(occupancy)
-        --* (1 / 5280) * 12))
-        --    as speed
-        (volume * 22) / nullifzero(occupancy) * (1 / 5280) * 12 as speed
+        case
+            when
+                --average occupancy should not be null and zero
```
britt-allen, line 71:
```suggestion
                -- average occupancy should not be null and zero
```
minor spacing suggestion for consistency

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -53,28 +54,27 @@ aggregated as (
...
+        sum(volume) as volume_sum, -- Sum of all the flow values
+        avg(occupancy) as average_occupancy, -- Average of all the occupancy values
+        -- weighted speed provides more accurate aggregation than average speed, therefore we will weighted speed
+        sum(volume * speed) / nullifzero(sum(volume)) as weighted_speed
     from station_raw
     group by id, lane, sample_date, sample_timestamp_trunc
 ),
 
-aggregated_speed as (
+-- select * from aggregated
```
britt-allen, line 65: this commented out code can be deleted

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -53,28 +54,27 @@ aggregated as (
         lane,
         count_if(volume is not null and occupancy is not null)
             as sample_ct, --Number of raw data samples
-        sum(volume) as volume, -- Sum of all the flow values
-        avg(occupancy) as occupancy -- Average of all the occupancy values
-        -- avg(speed) as speed_raw -- This code could be used if we use
-        -- actual speeds reported from raw data
+        sum(volume) as volume_sum, -- Sum of all the flow values
+        avg(occupancy) as average_occupancy, -- Average of all the occupancy values
```
britt-allen, line 58:
```suggestion
        -- Average of all the occupancy values
        avg(occupancy) as average_occupancy,
```

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -53,28 +54,27 @@ aggregated as (
         lane,
         count_if(volume is not null and occupancy is not null)
             as sample_ct, --Number of raw data samples
-        sum(volume) as volume, -- Sum of all the flow values
-        avg(occupancy) as occupancy -- Average of all the occupancy values
-        -- avg(speed) as speed_raw -- This code could be used if we use
-        -- actual speeds reported from raw data
+        sum(volume) as volume_sum, -- Sum of all the flow values
```
britt-allen, line 57:
```suggestion
        -- Sum of all the flow values
        sum(volume) as volume_sum,
```

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -53,28 +54,27 @@ aggregated as (
         lane,
         count_if(volume is not null and occupancy is not null)
             as sample_ct, --Number of raw data samples
```
britt-allen, line 56:
```suggestion
            -- Number of raw data samples
            as sample_ct,
```

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -53,28 +54,27 @@ aggregated as (
...
-        sum(volume) as volume, -- Sum of all the flow values
-        avg(occupancy) as occupancy -- Average of all the occupancy values
-        -- avg(speed) as speed_raw -- This code could be used if we use
-        -- actual speeds reported from raw data
+        sum(volume) as volume_sum, -- Sum of all the flow values
+        avg(occupancy) as average_occupancy, -- Average of all the occupancy values
+        -- weighted speed provides more accurate aggregation than average speed, therefore we will weighted speed
+        sum(volume * speed) / nullifzero(sum(volume)) as weighted_speed
     from station_raw
```
britt-allen, line 61: There is currently no model that is just named `station_raw`. You'll need to use the [dbt ref() function](https://docs.getdbt.com/reference/dbt-jinja-functions/ref) `{{ ref ('model_name') }}` replacing model name with the proper name of the model you want to reference.

britt-allen (commented):
Great work so far @mmmiah! This just needs a few more improvements. I made some suggestions which you can commit directly to your PR rather than duplicating the work in dbt Cloud.

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
mmmiah: Hi Brittany, Thank you for your quick review and feedback. the very top CTE is the station_raw.

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
mmmiah: Sure, dropped

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
britt-allen: Oh haha my mistake, thank you for pointing that out!

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -45,36 +45,38 @@ with station_raw as (
...
 ),
 
+
 aggregated as (
     select
         id,
         sample_date,
         sample_timestamp_trunc as sample_timestamp,
         lane,
         count_if(volume is not null and occupancy is not null)
```
britt-allen, line 55:
```suggestion
-- Number of raw data samples
        count_if(volume is not null and occupancy is not null)
```

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -45,36 +45,39 @@ with station_raw as (
...
-        (volume * 22) / nullifzero(occupancy) * (1 / 5280) * 12 as speed
+        case
+            when
+                -- average occupancy should not be null and zero
+                -- sum of the volume should not be null and zero
+                (average_occupancy is not null and average_occupancy != 0)
+                and (volume_sum is not null and volume_sum != 0)
+                then
+                    weighted_speed
+        end as five_mins_speed
```
ian-r-rose, line 79: Do we need this case statement at all? I feel like the formula above already does what you are looking for. This case statement may not be worth the expense of evaluating it.

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -45,36 +45,39 @@ with station_raw as (
...
-        avg(occupancy) as occupancy -- Average of all the occupancy values
-        -- avg(speed) as speed_raw -- This code could be used if we use
-        -- actual speeds reported from raw data
+            as sample_ct,
+        -- Sum of all the flow values
+        sum(volume) as volume_sum,
+        -- Average of all the occupancy values
+        avg(occupancy) as average_occupancy,
+        -- weighted speed provides more accurate aggregation than average speed, therefore we will weighted speed
+        sum(volume * speed) / nullifzero(sum(volume)) as weighted_speed
```
ian-r-rose, line 63: I was under the impression from our discussion in #165 that we wanted to actually report the speed as reported by the device. Were there further conversations/decisions about this? Are we sure we want to do *only* the weighted speed here, rather than reporting both speed and weighted speed?

ian-r-rose (commented):
Thanks @mmmiah! A couple of minor questions about the implementation

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
mmmiah: @ian-r-rose ! Great catch! Totally agree, I will drop this code!

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
mmmiah: @ian-r-rose , I tested three different speed scenarios like average, median and weighted speed for five minutes speed aggregation. It was found that mean speed is badly impacted by zero speed while median and weighted average speed are close to each others, however, weighted speed was selected as it is risk free from the impact of zero speed. For example if you have 10 different speeds in 5 mins (2 speed data in each mins) and six speed of this 10 are 0, then your median is going to be impacted by zero speed. To avoid this kind of situation, we consider weighted speed as it will avoid such kind of impacts. I and Ken did some comparison and realized that weighted speed is much more appropriate representation as some of the lanes may not have detectors and report zeros, this type of scenarios is not going to impact the weighted speed. However, we are open to have more discussion if something better is found. Thank you!

kengodleskidot:
@mmmiah it looks like you are making similar updates to the code in PR #196. If that is the case, you can close out this PR and include a reference in your close out comment to the new PR.

mmmiah:
oppps! Alright!

PR #189, opened 2024-04-25
changed instances of "pems" in model names to "diagnostics" renamed "detector_health_diag" schema and folder to "diagnostics"

### LINKED ISSUE #162: Rename `int_pems__` models
A number of models have been made with the `int_pems__` prefix. I think it might be a good idea to make them a bit more specific, since effectively every model in this project is related to PeMS. Per the [best practice](https://docs.getdbt.com/best-practices/how-we-structure/3-intermediate#intermediate-files-and-folders) documentation in dbt, the prefix could be something like line of business. In our case, I'd interpret that as broad use cases like diagnostics, performance metrics, etc.

What if we instead called a bunch of them `int_diagnostics__` or similar?

britt-allen:
Fixes #162


britt-allen:
This is ready for review and will hopefully be quick to review.

ian-r-rose (approved):
This looks good to me @britt-allen. Since some of these tables are quite expensive to recompute, would you be able to go into Snowflake and rename the tables and schemas manually once this is merged? Then they won't need to be recomputed from scratch.

britt-allen:
> This looks good to me @britt-allen. Since some of these tables are quite expensive to recompute, would you be able to go into Snowflake and rename the tables and schemas manually once this is merged? Then they won't need to be recomputed from scratch.

Good call, I can do that

britt-allen:
Looks like editing is not possible :/ @ian-r-rose

PR #192, opened 2024-04-25
Delay performance metric

### LINKED ISSUE #153: Delay
The Delay performance metric is the amount of additional time spent by the vehicles on a section of road due to congestion. This is the difference between the travel time at a non-congestion speed and the current speed. The congestion, or threshold, speed is usually 35MPH but in PeMS we compute the delay for a number of different thresholds so that we can accommodate different definitions of delay. The threshold values used are 35, 40, 45, 50, 55, and 60 MPH. The formula for this is simply D = F * (TT - TT_t) = F * (L/V - L/V_t). In these formulas, D is delay, TT is the travel time, TT_t is the travel time at the threshold speed, F is the flow, L is the length of the segment, V is the current speed and V_t is the threshold speed. Note that by definition, delay can never be negative.

We compute Delay for the various threshold values at the 5-minute level for each loop detector location (lane) as well as across all lanes at a given station. The measure of Delay can be summed up across time and space. This makes it quite easy for us to aggregate these measures to larger geographical regions and longer temporal periods.

britt-allen:
This will close #153


britt-allen:
I could use some help with understanding what's failing in sql fluff? I have tried running this locally and it has been running for over an hour.

ian-r-rose:
> I could use some help with understanding what's failing in sql fluff? I have tried running this locally and it has been running for over an hour.

I see some formatting-related issues in the logs: https://github.com/cagov/caldata-mdsa-caltrans-pems/actions/runs/8884558263/job/24393903279#step:9:366

If it's taking much longer to run in CI than in your development environment, there may a difference in how you are cloning an incremental model.

britt-allen:
tagging @tiffanychu90 for review

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -5,6 +5,8 @@
     snowflake_warehouse = get_snowflake_refresh_warehouse(small="XL")
 ) }}
 
+{% set V_t_list = var('V_t') %}
```
ian-r-rose, line 8: Is it necessary to set a local variable here rather than just using `var('V_t')` below?

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -62,6 +64,22 @@ vmt_vht_metrics as (
...
     from five_minute_agg_with_station_meta
+),
+
+delay_metrics as (
+    select
+        vvm.*,
+        /*  The formula for delay is: F * (L/V - L/V_t). F = flow (volume),
+        L = length of the segment, V = current speed, and V_t = threshold speed. */
+        {% for value in var("V_t") %}
+            vvm.volume * (vvm.length / nullifzero(vvm.speed)) - (vvm.length / {{ value }}) as delay_{{ value }}_mph
```
ian-r-rose, line 75: I notice that in the definition it says that "delay can never be negative". But this can easily be negative if the speed is greater than the given value in `V_t`. Is that intentional? Should this be wrapped in a `greatest()` function?

ian-r-rose (commented):
This is looking good to me. Just one question about the interpretation of "delay"

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
britt-allen: idts

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
britt-allen: by "this" do you mean the entire formula that calculates delay?

ian-r-rose: Yes, I recommend plugging in some test values to see how it behaves. But this is the type of thing that can cause problems down the line: if it's clearly documented that it can never be negative, and is in fact frequently negative, downstream metrics will probably break.

Probably worth a test in dbt!

britt-allen: I also don't think we'd want to only take the greatest delay, I think for every row and every `V_t` value we want to know delay

ian-r-rose: Not the greatest delay, that would be the `max` aggregation function. `greatest` can be used to clamp values (https://docs.snowflake.com/en/sql-reference/functions/greatest)

britt-allen:
@kengodleskidot can you chime in here with some thoughts on how to treat instances of negative delay? also, do we want one delay metric or delay calculated for every threshold (e.g. delay_35_mph, delay_40_mph, delay_45_mph, delay_50_mph, delay_55_mph, delay_60_mph)?

britt-allen:
> @kengodleskidot can you chime in here with some thoughts on how to treat instances of negative delay? also, do we want one delay metric or delay calculated for every threshold (e.g. delay_35_mph, delay_40_mph, delay_45_mph, delay_50_mph, delay_55_mph, delay_60_mph)?

Ken answered this in a call. We want to calculate for every threshold and we want to convert negative delay to 0

ian-r-rose:
> > @kengodleskidot can you chime in here with some thoughts on how to treat instances of negative delay? also, do we want one delay metric or delay calculated for every threshold (e.g. delay_35_mph, delay_40_mph, delay_45_mph, delay_50_mph, delay_55_mph, delay_60_mph)?
> 
> Ken answered this in a call. We want to calculate for every threshold and we want to convert negative delay to 0

Great! In that case I think that you do indeed want to wrap these calls in a `greatest()` call.

britt-allen:
> > > @kengodleskidot can you chime in here with some thoughts on how to treat instances of negative delay? also, do we want one delay metric or delay calculated for every threshold (e.g. delay_35_mph, delay_40_mph, delay_45_mph, delay_50_mph, delay_55_mph, delay_60_mph)?
> > 
> > 
> > Ken answered this in a call. We want to calculate for every threshold and we want to convert negative delay to 0
> 
> Great! In that case I think that you do indeed want to wrap these calls in a `greatest()` call.

I am not understanding why I'd want to do that. Can you explain, please?

britt-allen:
In dbt Cloud my table is building with the correct columns, in CI it is not. Spinning my wheels trying to figure out why that is. I  keep dropping the tables and that doesn't fix the issue.

kengodleskidot:
> In dbt Cloud my table is building with the correct columns, in CI it is not. Spinning my wheels trying to figure out why that is. I keep dropping the tables and that doesn't fix the issue.

@britt-allen I was able to run the performance model with no issue in dbt as well. There is a calculation issue I will leave a note for in the code. Once the code is updated I can run the model again in dbt and then we can check the CI. I believe you can eliminate the need for the test if you use the greatest function as previously mentioned. For the delay calculation the greatest function would look like "greatest(delay_calculation, 0)" that way if the delay calculation results in a negative value 0 would be returned since it would be the greatest of the two results. No need to perform any tests or checks. Here is the greatest function link: https://docs.snowflake.com/en/sql-reference/functions/greatest and let me know if you need anything else.

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
```diff
@@ -62,6 +62,25 @@ vmt_vht_metrics as (
...
+),
+
+delay_metrics as (
+    select
+        vvm.*,
+        /*  The formula for delay is: F * (L/V - L/V_t). F = flow (volume),
+        L = length of the segment, V = current speed, and V_t = threshold speed. */
+        {% for value in var("V_t") %}
+            iff(
+                vvm.volume * (vvm.length / nullifzero(vvm.speed)) - (vvm.length / {{ value }}) < 0, 0,
```
kengodleskidot, line 74: You are missing parentheses; the calculation should look like the following:
vvm.volume * ((vvm.length / nullifzero(vvm.speed)) - (vvm.length / {{ value }})) 
The volume is currently being multiplied to (length/speed) only instead of (length/speed - length/threshold_speed). A good rule of thumb is that there will only be delay if the actual speed travelled is less than the threshold speed. If you are travelling faster than the threshold speed, there is no delay.

Also, if you use the greatest function here you can simplify the code.

transform/tests/assert_delay_metrics_are_positive.sql
kengodleskidot, line 1: Nice data test!

transform/dbt_project.yml
```diff
@@ -31,6 +31,7 @@ vars:
   dev_model_look_back: -7
   day_start: "'05:00:00'"
   day_end: "'21:59:59'"
+  V_t: [35, 40, 45, 50, 55, 60]
```
kengodleskidot, line 34: We will use these threshold speeds for Productivity too!

kengodleskidot (dismissed):
Nice job! Once the calculation update is made let me know so I can do some QA/QC testing 😊

kengodleskidot:
@mmmiah based on my review of dbt Cloud error the issue may be with the updated column name's effect on the constant occupancy model that @britt-allen did. Make sure to loop her into this review and maybe setup a quick meeting so we can get this PR completed.

mmmiah:
@britt-allen are you availabe sometimes today? Thank you!

britt-allen:
I can meet for 30 minutes at 1:30 or 3:30, which works for you?

mmmiah:
Ken, does any of these time (1.30pm and 3.30 pm) works for you!

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
britt-allen: Thanks for the flag on the parentheses!

transform/dbt_project.yml
britt-allen: awesome! do you recommend we change the variable name?

britt-allen:
> > In dbt Cloud my table is building with the correct columns, in CI it is not. Spinning my wheels trying to figure out why that is. I keep dropping the tables and that doesn't fix the issue.
> 
> @britt-allen I was able to run the performance model with no issue in dbt as well. There is a calculation issue I will leave a note for in the code. Once the code is updated I can run the model again in dbt and then we can check the CI. I believe you can eliminate the need for the test if you use the greatest function as previously mentioned. For the delay calculation the greatest function would look like "greatest(delay_calculation, 0)" that way if the delay calculation results in a negative value 0 would be returned since it would be the greatest of the two results. No need to perform any tests or checks. Here is the greatest function link: https://docs.snowflake.com/en/sql-reference/functions/greatest and let me know if you need anything else.

Thank you for that explanation. When I read the greatest documentation twice earlier this week my understanding was not quite there. Technical docs are not always written with the greatest amount of empathy.

britt-allen:
This ready for final review and can hopefully be merged @kengodleskidot

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
mmmiah, line 1: Do we know the Unit of L (mile or m?), V amd V_t  (mph?) and also flow, F (sec or min?)? It will be great if we can just add the units for better understanding. Thank you!

mmmiah (approved):
I just finished my review and the code looks good to me! provided a minor comments to add unit of delay equation parameters within the comments for clarity. After Ken review, I think that we can merge this! Thanks!

transform/models/intermediate/performance/int_performance__five_min_perform_metrics.sql
kengodleskidot: L is in miles, V and V_t is mph, flow is veh/timeframe where the timeframe is typically 30 seconds or 5 minutes currently depending on the model it is being aggregated in.

kengodleskidot (approved):
Great job @britt-allen! The QA/QC on the delay values and the station/row count numbers look good.

PR #197, opened 2024-05-03
Update get snowflake refresh warehouse model/macro

kengodleskidot:
Updated lines 9 and 11 to use the variable names big and small instead of the warehouse size strings "4XL" and "XS"


britt-allen (approved):
gracias! :D

ian-r-rose:
Thanks @kengodleskidot!

PR #198, opened 2024-05-06
Added pems_clearinghouse_start_date variable

kengodleskidot:
Variable added to dbt_project.yml and staging__station_raw model to minimize computation resources while we continue to develop/refine our downstream models. The variable is currently not added to station_meta and station_status based models due to their relatively small size and limited equipment updates by some districts (primarily district 11). This is associated with #181.


transform/dbt_project.yml
mmmiah, line 1: the logic seems good to me!

mmmiah (approved):
The applied logic make sense! Thanks!

britt-allen (approved):
This will be good to go in once CI is passing. If you make further edits while this is still a draft please re-request review. Thank you @kengodleskidot!

kengodleskidot:
This PR closes #181.

kengodleskidot:
> This will be good to go in once CI is passing. If you make further edits while this is still a draft please re-request review. Thank you @kengodleskidot!

@britt-allen we can discuss which other models may need to add this variable. My thought is that applying the variable to the staging_clearinghouse__station_raw model will limit the data being reprocessed for all associated downstream models but if you can think of any other models this would be good to implement with let me know.

ian-r-rose:
This looks good to me, thanks @kengodleskidot! I agree that `stg_clearinghouse__station_raw` is the most important one to apply this to.

Since it's been in place for a week, have you noticed any improvements in development iteration times?

Issue #170, opened 2024-04-18
Imputation: compute from global coefficients


ian-r-rose:
Perform imputation on the 5-minute unimputed data based on linear regression from all detectors in a district, as documented in the [User manual](https://pems.dot.ca.gov/Papers/PeMS_Intro_User_Guide_v6.pdf).

There is also an opportunity to refine what is considered "global". What's described in the user manual is using all detectors in a district, but we could also do, e.g., all detectors in a district on the same freeway, or all detectors within a certain distance of the one in question.

<img width="772" alt="image" src="https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/5728311/deeba248-ad2c-4ead-80a2-bb60c1f4c8b7">

The algorithm is described in more detail [here](https://pems.dot.ca.gov/Papers/loops_asv2.pdf).

Additional Information: Linear regression from neighbors based on global coefficients. This imputation method is a method similar to compute from local coefficients. The only caveat is that there are some loops which never report data. In that situation it is not possible to compute local regression coefficients. Therefore, we compute global regression coefficients. These coefficients represent the general relationship seen throughout the route.

mmmiah:
We are working on this and have opened new issues #199

britt-allen:
Why is this issue closed? You linked to a pull request, not a new issue @mmmiah . I don't understand how pull request 199 relates to this issue. This issue can't be closed unless a PR closes it out or we decide as a team this is unplanned work.

mmmiah:
@britt-allen , alright, the pull #199 aims to fulfill the goal mentioned in this issue, but the PR #199 is in initial stage. I thought/misunderstand that you want me to close #170 based on your comments. I have edited the final goal of #199 as 'This code identified all the upstream and downstream stations within five miles buffer that is within same district, route, directions and highways. The goal of this model is to identify the stations that can be used to develop linear regression model and determine the global coefficient to impute the missing performance metrics where local regression models fails.'

junlee-analytica:
Global coefficients changed to regional coefficients.

britt-allen:
Is this issue still being worked on?

ian-r-rose:
Yes, in #250

PR #199, opened 2024-05-06
Global_Regression_for_Imputation

mmmiah:
This code identified all the upstream and downstream stations  within five miles buffer that is within same district, route, directions and highways. The goal of this model is to identify the stations that can be used to develop linear regression model and determine the global coefficient to impute the missing performance metrics where local regression models fails. The details are described in issue #170


transform/models/intermediate/clearinghouse/int_cleaninghouse_global_nearby_station.sql
```diff
@@ -0,0 +1,45 @@
...
+    -- e.g., on/off ramps. So it makes sense to restrict these comparisons to some
+    -- types. Is this the right set?
+    where a.type in ('HV', 'ML', 'CH', 'CD')
+),
+
+nearest_stations_within_five_miles as (
+    select *
+    from station_pairs
+    -- select the stations those are within 5 miles upstream and downstream of the station
+    where delta_postmile >= -5.0 and delta_postmile <= 5.0 and id != other_id
```
kengodleskidot, line 29: This CTE can be eliminated by adding the "where" conditions you have here into the previous CTE. You may also want to look at using the ABS function (https://docs.snowflake.com/en/sql-reference/functions/abs) to simplify the delta_postmile where statement...abs(delta_postmile) <= 5.0

transform/models/intermediate/clearinghouse/int_cleaninghouse_global_nearby_station.sql
```diff
@@ -0,0 +1,45 @@
...
+
+nearest_stations_within_five_miles_count as (
+    select
+        id,
+        other_id,
+        district,
+        freeway,
+        direction,
+        type,
+        delta_postmile,
```
kengodleskidot, line 40: The first seven columns in your select statement can be called by using the "*," statement to help simplify your code.

kengodleskidot:
This is a good start @mmmiah! I am not sure how the station proximity count solves our global stations imputation issue (#170) so I am wondering if this should be a draft PR. This seems like one of the first models to be developed as part of the bigger global station solution. Let us know what the end goal is for this version of imputation you are working on.

kengodleskidot (commented):
This is a good start @mmmiah! I am not sure how the station proximity count solves our global stations imputation issue (https://github.com/cagov/caldata-mdsa-caltrans-pems/issues/170) so I am wondering if this should be a draft PR. This seems like one of the first models to be developed as part of the bigger global station solution. Let us know what the end goal is for this version of imputation you are working on.

transform/models/intermediate/clearinghouse/int_cleaninghouse_global_nearby_station.sql
mmmiah: alright!

transform/models/intermediate/clearinghouse/int_cleaninghouse_global_nearby_station.sql
mmmiah: great suggestion! just implemented!

transform/models/intermediate/clearinghouse/int_cleaninghouse_global_nearby_station.sql
```diff
@@ -0,0 +1,45 @@
...
+        -- TODO: is this the best metric for "nearest"? State postmiles often have
+        -- some string part to them, making math a little trickier.
+        b.absolute_postmile - a.absolute_postmile as delta_postmile
+    from station_meta as a
+    inner join station_meta as b
+        on a.freeway = b.freeway and a.direction = b.direction and a.type = b.type and a.district = b.district
+    -- TODO: distance comparisons don't seem appropriate for all station types,
+    -- e.g., on/off ramps. So it makes sense to restrict these comparisons to some
+    -- types. Is this the right set?
+    where a.type in ('HV', 'ML', 'CH', 'CD')
```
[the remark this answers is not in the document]
mmmiah: dropped, CH and CD for now! Thanks for catching it!

britt-allen:
> #170

I agree with Ken and would love to understand the end goal.

britt-allen:
Is this model aimed to close issue #170 or something else? I am unclear on the goal of this. Please link the appropriate issue when you get a chance @mmmiah

britt-allen:
Is there an existing issue that captures the current goal of this current code and not future goals? I am still a little shaky on how this all fits into the bigger picture @kengodleskidot @mmmiah

transform/models/intermediate/clearinghouse/int_imputation_global_regression_coefficients.sql
```diff
@@ -0,0 +1,91 @@
...
+        detector_id,
+        station_id,
+        district,
+        len(lane_number) as lane
+    from {{ ref('int_clearinghouse__most_recent_station_status') }}
+    where detector_status = 'operational'
+),
+
+station_counts as (
+    select * from {{ ref('int_performance__five_min_perform_metrics') }}
```
ian-r-rose, line 15: Is there a reason why you are selecting from `int_performance__five_min_perform_metrics` rather than `int_clearinghouse__five_minute_station_agg`? It doesn't look like you are using any of the performance metrics here, so I don't think introducing a coupling with performance metrics is necessary.

From a high level, we intend to eventually compute the performance metrics *from* the imputed data. Selecting directly from the unimputed data would probably save us a circular dependency down the line.

transform/models/intermediate/clearinghouse/int_imputation_global_regression_coefficients.sql
```diff
@@ -0,0 +1,91 @@
...
+    where sample_date = current_date - interval '5 day'
```
ian-r-rose, line 16: Pretty soon we'll want to remove all references to `current_date` and ensure these models are idempotent. I think this is fine for a work-in-progress, but ideally a full refresh of a model will not depend on the day that it's been built on.

transform/models/intermediate/clearinghouse/int_cleaninghouse_global_nearby_station.sql
ian-r-rose, line 1: Per our naming conventions, this should be something like `int_clearinghouse__global_nearby_stations`

transform/models/intermediate/clearinghouse/int_imputation_global_regression_coefficients.sql
```diff
@@ -0,0 +1,91 @@
...
+-- now it seems that some of the station within the buffer have missing volume and occupancy
+-- we will drop all the Null volumns and occupancy before developing multiple linear regression for the day
+cleaned_model_data as (
+    select *
+    from station_counts_pairwise
+    where
+        (volume is not null and other_volume is not null)
+        and (occupancy is not null and other_occupancy is not null)
+        and (speed is not null and other_speed is not null)
+),
```
ian-r-rose, line 68: It's probably more efficient to do this filtering *before* the expensive join in `station_counts_pairwise`.

transform/models/intermediate/clearinghouse/int_global_regression_deployment.sql
```diff
@@ -0,0 +1,102 @@
+{{ config(materialized='table') }}
+
+--  lets consider the unimputed dataframe
+with unimputed as (
+    select * from {{ ref('int_performance__five_min_perform_metrics') }}
+    where sample_date = dateadd(day, -5, current_date)
+),
```
ian-r-rose, line 7: Same here, I don't think we want to select from the performance metrics table

transform/models/intermediate/clearinghouse/int_global_regression_deployment.sql
```diff
@@ -0,0 +1,102 @@
...
+        sample_date,
+        sample_timestamp,
+        volume_sum,
+        occupancy_avg,
+        speed_five_mins
+    from unimputed
+    -- when volume and occupancy is present we already used the formula to calculate the speed
+    -- here we will impute where all three has null value
+    -- however, we will not impute where volime zero, occupancy zero and speed is null
+    where volume_sum is null and occupancy_avg is null and speed_five_mins is null
```
ian-r-rose, line 28: Do we also want to impute if some of these are null but not others? I think we might...

transform/models/intermediate/clearinghouse/int_global_regression_deployment.sql
```diff
@@ -0,0 +1,102 @@
...
+        id,
+        lane,
+        sample_date,
+        sample_timestamp,
+        volume_sum,
+        occupancy_avg,
+        speed_five_mins,
+        false as is_imputed
+    from unimputed
+    where volume_sum is not null and occupancy_avg is not null
```
ian-r-rose, line 92: Whatever condition we have for selecting out the missing ones, it should be the inverse for the non-missing ones, otherwise we'll end up with gaps or duplicates. It might even be worth setting a jinja variable for the expression to enforce it.

ian-r-rose (commented):
Thanks @mmmiah!

I wonder if we should rename this approach "regional" imputation rather than "global"? That is a more descriptive name to my mind.

Given how similar this is to the local version, eventually I think it would make sense to combine them (a "local" version is just a "regional" one with a more restrictive definition of neighbors). But I'd be comfortable doing that later.

mmmiah (commented):
This is great feedbacks. I am working on this  and will reflect this change soon! I have just tested and found that previous logic to separate whether we need imputation or not, had the hole as you suspected. I have revised that code. Now I am testing why the model is imputing really low number of stations. We have around 39 million ids for five minutes, out of these 39 millions data 29 millions have missing data and another 10 millions are the good data. Now if I consider 5 miles buffer, model imputes only 2100 rows, for 10 miles buffer it imputes 4484 data points. I will work on this and will update on what sizes of buffer we need or we need to consider entire districts. Thanks!

transform/models/intermediate/clearinghouse/int_imputation_global_regression_coefficients.sql
ian-r-rose: I think it's totally fine to calculate the five minute speed formula on the fly, as it's a very lightweight transformation (if I understand correctly). If that formula gets used in a number of different places, it could also be a good case for a view or ephemeral materialization.

transform/models/intermediate/clearinghouse/int_global_regression_deployment.sql
ian-r-rose: I seem to recall seeing some values where some were null but not others (it struck me as odd too). But I might be mis-remembering.

transform/models/intermediate/clearinghouse/int_imputation_global_regression_coefficients.sql
```diff
@@ -0,0 +1,91 @@
...
+
+-- identify the station that is operational
+with station_status as (
+    select
+        detector_id,
+        station_id,
+        district,
+        len(lane_number) as lane
+    from {{ ref('int_clearinghouse__most_recent_station_status') }}
+    where detector_status = 'operational'
```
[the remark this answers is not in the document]
kengodleskidot: This issue with using int_diagnostics__detector_status is that the status is determined only after a full days' worth of data has been analyzed. For the imputations, we are calculating missing values for flow, occupancy and speed the same day multiple times. You will not have the current days status until the next day thus the current days stations are not in that model. I created the int_clearinghouse__active_stations model which is a relatively small and lightweight model which contains all active stations for any day including the current day.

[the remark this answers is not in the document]
mmmiah: Thanks@ ken. I have incorporated you suggestions in the latest code!

transform/models/intermediate/clearinghouse/int_imputation_global_regression_coefficients.sql
mmmiah: @ian-r-rose , great! I considered five mins agg file and calculated the five mins speed . Updated codes are up for the re-review

mmmiah:
@ian-r-rose .I am going to start updating the regional and global regression while taking local regression as the foundation. Based on last discussion, I think that I am going to work directly on your branch nearby-detectors! 

Here is the pan:
a. For regional regression, we will keep everything same as local regression while increase the upstream and downstream coverage ring buffer 5 miles.
b. For global regression we will consider all types of the road within district.
c. Finally we will merge regional regression and global regression estimate in local regression table.

a short meeting tomorrow would be a helpful before I  start updating these two models to understand from where I should start updating these two models! Let me know the time works for you tomorrow! Thank you!

ian-r-rose:
Sounds great!

> @ian-r-rose .I am going to start updating the regional and global regression while taking local regression as the foundation. Based on last discussion, I think that I am going to work directly on your branch nearby-detectors!

I think it would be better to just create a new branch from `main`, now that `nearby-detectors` is merged. This should help create fewer merge conflicts down the line.

> Here is the pan: a. For regional regression, we will keep everything same as local regression while increase the upstream and downstream coverage ring buffer 5 miles. b. For global regression we will consider all types of the road within district. c. Finally we will merge regional regression and global regression estimate in local regression table.

> a short meeting tomorrow would be a helpful before I start updating these two models to understand from where I should start updating these two models! Let me know the time works for you tomorrow! Thank you!

This sounds like a great plan to me! I could meet tomorrow during the following windows:

- 9-10:30
- 12-1
- 2-5

ian-r-rose:
Based on the discussions we've had with @kengodleskidot, it looks like the global regression is based not on station pairs, but on a regression between each station and some district-wide average. Which is good, because I think that looking at every station pair in the whole district-freeway combo might be too expensive!

So I think we'll want to take a different approach here to calculating the regression coefficients that doesn't rely on station pairs or nearby stations, but instead looks at a district-freeway aggregate. Does that make sense @mmmiah, or do you have a different understanding?

I have some experiments along those lines that I can open as a draft for discussion.

PR #203, opened 2024-05-09
Productivity

kengodleskidot:
Develop max capacity model or station/lanes and add lost productivity calculation to performance metrics.


transform/models/intermediate/performance/int_performance__max_capacity.sql
mmmiah, line 1: Line 48 did not compare apple to apple, volume sum is for 15 mins while 173 v/l is for 5 minutes. This comparison is inconsistent. Lets have discussion with PeMS people what is actual logic behind of doing this or if they have any alternative suggestions???

Other codes look good to me

mmmiah (commented):
Need to come up with conclusion about how to calculate 5mins capacity after discussing with PeMS people.

mmmiah (commented):
edited the comment

kengodleskidot:
@britt-allen I believe this CI continues to fail because the dbt_cloud_pr_15_203_clearinghouse.int_clearinghouse__five_minute_station_agg model is being built with the old column naming and not the updated one (i.e. using VOLUME as the column name instead of VOLUME_SUM). Let me know when you have time to go over and troubleshoot this 😊

transform/models/intermediate/performance/int_performance__max_capacity.sql
kengodleskidot: The first CTE gets the 5-minute aggregate station and volume data, the second CTE adds up volume over three 5-minute intervals (15 minutes total) per station/lane/date, the third CTE returns the greatest of the maximum 15-minute volume total per station/lane or 173 v/l over 5 minutes.

I agree with your first point and we can divide the 15-minute maximum flow by 3 to get a 5-minute flow but this will result in negative productivity. Since the 15-minute maximum flow is what is stated in PeMS that is what the code currently has but I agree that we should discuss this with the Technical Committee members to get their understanding.

britt-allen:
> @britt-allen I believe this CI continues to fail because the dbt_cloud_pr_15_203_clearinghouse.int_clearinghouse__five_minute_station_agg model is being built with the old column naming and not the updated one (i.e. using VOLUME as the column name instead of VOLUME_SUM). Let me know when you have time to go over and troubleshoot this 😊

I resolved this for now by dropping and rebuilding production tables. I'll do some research to see if there are other ways to resolve schema changes. I am in a meeting right now, but will review this PR right after!

transform/models/intermediate/performance/int_performance__max_capacity.sql
britt-allen: The answer to your second question is yes @mmmiah. Ken or I can walk you through that window function if helpful. We'll also hold a training session on window functions in the near future to get the team more comfortable with using and understanding them.

britt-allen (approved):
This looks great Ken, thank you!

transform/models/intermediate/performance/int_performance__max_capacity.sql
britt-allen: I am going to merge this for now, but not delete this branch so we can revisit this if needed after technical committee discussion. @kengodleskidot

Issue #169, opened 2024-04-18
Imputation: compute from local coefficients


ian-r-rose:
Perform imputation on the 5-minute unimputed data based on linear regression from a detector's neighbors (detectors in the same station, as well as those upstream and downstream), as documented in the [User manual](https://pems.dot.ca.gov/Papers/PeMS_Intro_User_Guide_v6.pdf).

<img width="772" alt="image" src="https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/5728311/f4e3b5ef-25d3-4c23-b49e-3a531cbeabad">

The algorithm is described in more detail [here](https://pems.dot.ca.gov/Papers/loops_asv2.pdf).

mmmiah:
@ian-r-rose @kengodleskidot This is the step suggested by imputation paper to impute volume and occupancy-
step 1: Consider 30 seconds raw data from 5 am to 10 Pm (17hrs)
Step 2: Identify the bad sensors using following criteria-
-(# of sample that  have occupancy=0) >1200 or
-(# of samples that have occupancy>0 and flow=0)>50 or
- (# of samples that have occupancy>0.35)>200 or 
- Entropy of occupancy sample <4

Step 3: Identify good sensor after separating the bad sensor
Step 4: Find two nearest good sensor for each bad sensor and develop the linear regression using five days historical data and apply the model to impute the bad detectors volume and occupancy.

Do we want to follow this or want to do some brain storming before starting of this imputation?? Thank you!

ian-r-rose:
Hi @mmmiah, let's meet about it this week! Do you have any availability tomorrow? I'd like to walk you through a work-in-progress branch that starts on this. It uses the detector status as evaluated using `int_pems__detector_status` to identify "bad" sensors, and uses a linear regression between "good" neighboring stations to impute volume and occupancy.

One of the tricks is going to be to decide how often to re-compute coefficients for the regression: it's a relatively expensive operation to compute them, but doesn't necessarily need to be done daily (it could, e.g., be done quarterly).

mmmiah:
@ian-r-rose ,
Hi Ian, yes,I am available between 10 am to 2.30 pm tomorrow. Please send me the invitation. Thank you!

junlee-analytica:
Local coefficients have been developed, preparing for review and deployment once Ian is back.

junlee-analytica:
[...]

PR #209, opened 2024-05-16
Local Regression for Imputation

mmmiah:
This is the pull followed by Issue #169 (please follow #169 to understand the local regression in more details) . This model originally build by Ian, later I just edited and modified some parts of this model. The goal of this model is to impute the missing volume, occupancy and speed (newly included) in the detector or lane level for each of station using just upstream and downstream data and develop simple linear regression. Once the model is developed, the model can be applied in daily level. The model parameters (slope and coefficient) was estimated daily level and applied at five minutes level in imputation phase. This code is ready for your review and feedback.


mmmiah:
@ian-r-rose @kengodleskidot  , We developed local and global regression just for 'ML' and 'HV'. Right now we applied it for all types of the road. I think that we want to be consistent with the application of model only for 'ML' and 'HV'.  Another thing is that I believe that we want to impute 'ML' and 'HV' missing data first with local regression and then the remaining gap with global regression or do we want to report this separately?

ian-r-rose:
> @ian-r-rose @kengodleskidot , We developed local and global regression just for 'ML' and 'HV'. Right now we applied it for all types of the road. I think that we want to be consistent with the application of model only for 'ML' and 'HV'.

I agree that we should be consistent. The initial reason I wanted to restrict myself to ML and HV was that it wasn't clear to me that neighboring stations made much sense for, e.g., a ramp. But if you can think of a way to do imputation for those other cases, we might be able to relax that restriction.

> Another thing is that I believe that we want to impute 'ML' and 'HV' missing data first with local regression and then the remaining gap with global regression or do we want to report this separately?

I think we should do the imputation in order, first local, then global/regional, then any other methods. Since these imputation methods are somewhat expensive, I don't think we necessarily want to compute them if we don't need to.

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+                    and unimputed.speed_five_mins is null
+                )
+                or (
+                    unimputed.volume_sum is null
+                    and unimputed.occupancy_avg is not null
+                    and unimputed.speed_five_mins is null
+                )
+                then 'yes'
+            else 'no'
+        end as is_imputation_required
```
ian-r-rose, line 42: I suspect this can be simplified a lot (to the point where a `case` statement is not necessary). If the goal is to impute when

1. All three are null
2. Either volume or occupancy is null

I think it will be covered by `(volume is null) or (occupancy is null)`, and having a boolean dtype. You have another case where you don't impute when occupancy is null, volume is null, and speed is _not_ null. But I suspect you might want to impute in that case as well.

In any event, I do thing doing boolean logic here would make things much simpler than a big case statement:
```suggestion
        (unimputed.volume_sum is null) or (unimputed.occupancy_avg is null) as imputation_required
```

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+        id,
+        lane,
+        -- sample_hour,
+        sample_date,
+        sample_timestamp,
+        volume_sum,
+        occupancy_avg,
+        speed_five_mins
+    from counts_with_imputation_status
+    where is_imputation_required = 'yes'
```
ian-r-rose, line 58: If you do the above and make it a boolean, this just becomes `where imputation_required`.

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+        lane,
+        sample_date,
+        sample_timestamp,
+        -- sample_hour,
+        volume_sum,
+        occupancy_avg,
+        speed_five_mins,
+        false as is_imputed
+    from counts_with_imputation_status
+    where is_imputation_required = 'no'
```
ian-r-rose, line 74: And here

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+with unimputed as (
+    select
+        id,
+        lane,
+        sample_date,
+        sample_timestamp,
+        volume_sum,
+        occupancy_avg,
+        speed_weighted,
+        extract(hour from sample_timestamp) as sample_hour,
```
ian-r-rose, line 12: What is this for?

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+                    and unimputed.speed_five_mins is null
+                )
+                then 'yes'
+            else 'no'
+        end as is_imputation_required
+    from unimputed
+),
+
+-- separate the dataframe that have missing volume,occupancy and speed
+missing_vol_occ_speed as (
```
ian-r-rose, line 47: I think this name is not fully accurate, why not something like `counts_requiring_imputation`?

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+            least(greatest(median(occupancy_slope * occupancy_avg_nbr + occupancy_intercept), 0), 1)
+        ) as occupancy_avg_imp,
+        -- Flag for occupancy imputation
+        coalesce(occupancy_avg is null, false) as is_imputed_occupancy,
+        -- Speed calculation
+        case
+            when
+                (speed_five_mins is null and volume_sum > 0 and occupancy_avg > 0)
+                or (speed_five_mins is null and volume_sum is null and occupancy_avg is null)
+                then least(greatest(median(speed_slope * speed_five_mins_nbr + speed_intercept), 0), 100)
```
ian-r-rose, line 139: I'm not sure I agree that we should be clamping to 100 mph here. If we think it's non-physical, I might want to try the next imputation method (and thus we'd maybe make it `null`?)

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+        sample_date,
+        sample_timestamp,
+        volume_sum,
+        occupancy_avg,
+        speed_weighted,
+        extract(hour from sample_timestamp) as sample_hour,
+        coalesce(speed_weighted, (volume_sum * 22) / nullifzero(occupancy_avg) * (1 / 5280) * 12)
+            as speed_five_mins
+    from {{ ref('int_clearinghouse__five_minute_station_agg') }}
+    where sample_date = dateadd(day, -3, current_date)
```
ian-r-rose, line 16: I'd like to take a look at getting rid of all the uses of `current_date`, except in the case of incremental logic.

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+        coalesce(
+            occupancy_avg,
+            least(greatest(median(occupancy_slope * occupancy_avg_nbr + occupancy_intercept), 0), 1)
+        ) as occupancy_avg_imp,
+        -- Flag for occupancy imputation
+        coalesce(occupancy_avg is null, false) as is_imputed_occupancy,
+        -- Speed calculation
+        case
+            when
+                (speed_five_mins is null and volume_sum > 0 and occupancy_avg > 0)
```
ian-r-rose, line 137: I'm surprised you want to impute the speed in this case. If volume and occupancy are available, can't we compute speed from them directly rather than imputing from neighbors?

ian-r-rose (commented):
Thanks @mmmiah, making good progress.

I would encourage you to try to simplify a lot of your `case` statements, I'm finding them difficult to read, and I suspect a lot of them are not necessary for the logic.

Do you mind if I continue pushing some commits here to work on having multiple measures of the coefficients going back in time? Overall, I'd like to get rid of all uses of `current_date` here so that the models can be re-run without depending on the date they are built on.

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
mmmiah: @ian-r-rose , sure, go ahead and replace this with enhanced code! Thank you!

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
mmmiah: 'extract(hour from sample_timestamp) as sample_hour,' this line of code is type, I believe it came from aggregation task. We do not need this hour

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
mmmiah: please feel free to replace the name as 'counts_requiring_imputation'

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
mmmiah: That is another approach, I and ken already discussed. we can decide what is best for us!

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,204 @@
...
+        end as is_imputed_occupancy,
+        case
+            when speed_five_mins_imp is not null then is_imputed_speed else false
+        end as is_imputed_speed
+    from missing_imputed_vol_occ_speed
+),
+
+-- combine imputed and non-imputed dataframe together
+-- -- combine imputed and non-imputed dataframe together
+local_regression_imputed_value as (
```
ian-r-rose, line 175: Musing a bit about this `union all`: I'm wondering if it's better to restructure this as a left join with the original unimputed table so that we can build up a table the results of multiple imputation methods in separate columns. This would help compare different methods, and a final step could be coalescing different methods to get our final imputed table.

My other motivation here is that the imputation methods are imperfect, and it will be difficult to get imputations for every missing value for a given method. Rather than trying to ensure we maintain all the nulls through a given imputation method, it may be easier to do a left join at the end.

I'm open to push-back here, what do you think @mmmiah?

mmmiah: I like the idea if you think that we keep the data frame neat and clean as much as possible with only needed column. Please go ahead and implement this idea! Thanks for doing all these improvement on coding!

transform/models/intermediate/diagnostics/int_diagnostics__real_detector_status.sql
```diff
@@ -0,0 +1,25 @@
...
+
+with detector_status as (
+    select * from {{ ref("int_diagnostics__detector_status") }}
+),
+
+detectors as (
+    select * from {{ ref("int_clearinghouse__station_status") }}
+),
+
+/* TODO: we should ensure that all of the detectors in the detector
```
britt-allen, line 11: is there an existing issue that captures this to-do and links back to this PR so we know to update this model when that one goes in?

britt-allen (commented):
This looks good to me. Great work @ian-r-rose and @mmmiah. @kengodleskidot this is pending your final review. please merge after review unless u have feedback or Qs

britt-allen (approved):

transform/models/intermediate/diagnostics/int_diagnostics__real_detector_status.sql
ian-r-rose: Good point, I've written it up in #226. It's also related to #217

mmmiah (commented):
looks good to me! Great job@ Ian for accomplishing this!

transform/models/intermediate/imputation/int_imputation__five_minute_aggregation.sql
```diff
@@ -0,0 +1,172 @@
...
+    select
+        id,
+        lane,
+        sample_date,
+        sample_timestamp,
+        volume_sum,
+        occupancy_avg,
+        speed_five_mins
+    from unimputed
+    where detector_is_good
```
kengodleskidot, line 90: We will want to explore how we fill in data holes for situations where samples are missing in the five minute model. If we have less than 10 thirty-second samples in our five_minute_station_agg model do we want to fill in that hole by normalizing the samples we do receive over give minutes or do we want to input imputed data to fill in that hole. If we do use normalization are we imputing based on the normalized data or the five_minute_agg data with holes?

ian-r-rose: Thanks @kengodleskidot, this is a great point. Can you open a new issue for it so it doesn't get lost?

PR #216, opened 2024-05-23
Sync all columns on schema changes

britt-allen:
added this flag: +on_schema_change: "sync_all_columns" so less manual full refreshes have to be kicked off. see dbt docs: https://docs.getdbt.com/docs/build/incremental-models#how-do-i-rebuild-an-incremental-model

### DIFF
--- a/transform/dbt_project.yml
+++ b/transform/dbt_project.yml
@@ -35,6 +35,7 @@ vars:
   pems_clearinghouse_start_date: "'2023-01-01'"
 
 models:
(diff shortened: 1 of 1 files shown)


britt-allen:
> I don't think this option does what you expect: per the docs:
> 
> > Note: None of the on_schema_change behaviors backfill values in old records for newly added columns. If you need to populate those values, we recommend running manual updates, or triggering a --full-refresh.

Ooof good flag, closing this!

Issue #186, opened 2024-04-25
Bottleneck Duration


jkarpen:
Bottleneck Identification Algorithm
The PeMS system runs a bottleneck identification algorithm every day. The original algorithm was presented in, "Systematic Identification of Freeway Bottlenecks," by Chen, C., Skabardonis, A., Varaiya, P. (2003). Transportation Research Board, 2004.

This algorithm attempts to identify bottlenecks on the freeway system at every detector. At a high level, a bottleneck exists at a particular detector when there is a persistent drop in speed from the current detector to the detector immediately upstream. Specifically we have the following defining conditions:

There is a drop in speed of at least 20 mph.
The speed at the current detector is less than 40 mph.
The detectors are less than 3 miles apart.
The speed drop persists for at least 5 out of any 7 contiguous 5-minute data points.
When localizing PeMS to metric units these thresholds are translated directly.
If all of these conditions are met then we declare that there is a bottleneck at this location that has been activated for all of the 7 5-minute time points (technically the bottleneck is between these two detectors - we don't know where exactly. We declare that the bottleneck is at the detector where the speed has dropped.) We perform this analysis for each of three time periods: AM shift (5am-10am), noon shift (10am-3pm), and PM shift (3pm-8pm). For each location where a bottleneck is activated we compute the following:

**Bottleneck Duration**: This is how long the bottleneck was active during that particular shift on that day.

britt-allen:
AM off peak (or day off peak??), AM peak (or Day peak??), PM peak, PM off peak, holiday - these are additional time frames we might need to consider? Can you clarify and expand on what you mentioned this morning @kengodleskidot 

We discussed potentially creating a table to capture this.

britt-allen:
Right now, the three time periods listed above are not something we need to worry about as identified in this morning's discussion. We'll re-visit this at a later time.

britt-allen:
Based on the technical discussion yesterday do we want to continue using delay_60_mph or switch to delay_65_mph? @kengodleskidot

britt-allen:
Here's the outline I've been using to keep me grounded in this work: https://docs.google.com/spreadsheets/d/1gv4S8BsSt4gScx8kHNzhGjSH-z4srUGEiehqmLdvfdg/edit#gid=0

kengodleskidot:
> Based on the technical discussion yesterday do we want to continue using delay_60_mph or switch to delay_65_mph? @kengodleskidot

[...]

PR #219, opened 2024-05-30
Bottleneck duration

britt-allen:
This will close issue #186


transform/models/intermediate/performance/int_performance__calculate_distance.sql
```diff
@@ -0,0 +1,37 @@
...
+station_meta as (
+    select
+        meta_date,
+        freeway,
+        direction,
+        type,
+        lanes as lane,
+        to_decimal(absolute_postmile, 6, 3) as absolute_postmile,
+        id
+    from {{ ref ("stg_clearinghouse__station_meta") }}
```
ian-r-rose, line 12: You'll want to select from the intermediate model, which adds the `_valid_from` and `_valid_to` fields.

transform/models/intermediate/performance/int_performance__calculate_distance.sql
```diff
@@ -0,0 +1,37 @@
...
+
+),
+
+calculate_distance_delta as (
+    select
+        *,
+        absolute_postmile
+        - lag(absolute_postmile)
+            over (partition by meta_date, freeway, direction, type, lane order by distance_rank)
+            as distance_delta
```
ian-r-rose, line 33: I don't understand why we need to split this into two window functions. They are partitioned by the same thing, so I would think you could just sort by absolute postmile in the first place

transform/models/intermediate/performance/int_performance__calculate_distance.sql
```diff
@@ -0,0 +1,37 @@
+with
+
+station_meta as (
+    select
+        meta_date,
+        freeway,
+        direction,
+        type,
+        lanes as lane,
+        to_decimal(absolute_postmile, 6, 3) as absolute_postmile,
```
ian-r-rose, line 10: Why do you need to convert to decimal?

britt-allen: I dunno

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,90 @@
...
+        sample_date,
+        sample_timestamp,
+        speed_five_mins,
+        delay_60_mph,
+        coalesce(speed_five_mins < 40, false) as speed_less_than_40
+
+    from {{ ref ("int_performance__five_min_perform_metrics") }}
+    where
+        to_time(sample_timestamp) >= {{ var("day_start") }}
+        and to_time(sample_timestamp) <= {{ var("day_end") }}
```
kengodleskidot, line 18: You can remove the timeframe criteria (day_start and day_end) from the where statement since we would want to identify bottlenecks that occur anytime during the day.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,90 @@
...
+            and sample_date > (
+                select
+                    DATEADD(
+                        day,
+                        {{ var("incremental_model_look_back") }},
+                        MAX(sample_date)
+                    )
+                from {{ this }}
+            )
+        {% endif %}
```
kengodleskidot, line 30: Since this block only applies to an incremental model you may want to update this to an incremental model. It is currently a table as defined on line 1.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,90 @@
...
+            and d._valid_to > f.sample_date
+            and d.id = f.id
+),
+
+calculate_speed_delta as (
+    select
+        *,
+        speed_five_mins
+        - lag(speed_five_mins)
+            over (partition by sample_timestamp, freeway, direction, type, lane order by absolute_postmile asc)
```
kengodleskidot, line 58: Since bottlenecks are analyzed across all lanes you will likely want to remove the lane as part of this partition.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,90 @@
...
+            over (
+                partition by sample_timestamp, freeway, direction, type, lane
+                order by absolute_postmile asc rows between 6 preceding and current row
+            )
+            as persistent_speed_drop
+
+    from bottleneck_criteria_check
+)
+
+select * from drop_persists_check
```
kengodleskidot, line 90: Great progress so far @britt-allen! We will want to incorporate the speed drop persists for at least 5 out of any 7 contiguous 5-minute data points criteria and ultimately, we want the duration, extent and delay associated with the bottlenecks. I have added the technical paper to #186 for reference.

kengodleskidot (commented):
Great progress @britt-allen! Let me know if you want to go over anything regarding this model.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,90 @@
...
+with
+
+five_minute_pm as (
+    select
+        id,
+        lane,
+        sample_date,
+        sample_timestamp,
+        speed_five_mins,
+        delay_60_mph,
```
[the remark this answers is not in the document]
britt-allen: Once Mintu's PR is merged I'll swap values

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: I actually misspoke. I want to materialized the performance metric model incrementally, but not the bottleneck model

britt-allen: I actually want to materialize this incrementally

britt-allen:
@kengodleskidot we noticed a few instances where speed was 0 and triggered a bottleneck_start because the other conditions were true as well. do we want to filter out instances of speed being zero for flagging the start of a bottleneck?

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: Can you clarify if for this PR we just want to focus on getting bottleneck start and once tackled, I can open a new pr for bottleneck continues and ends respectively? Ian and I seemed to remember different things about the intent for this PR based on our oral disscusion last week.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,117 @@
...
+                speed_weighted < 40
+                and abs(distance_delta_sw) < 3
+                and speed_delta_sw <= -20
+                and speed_prev_sw >= 40
+                and (direction = 'S' or direction = 'W')
+                then 1
+            else 0
+        end as bottleneck_start
+        /** this criteria below needs to remain true for five out of seven 5-minute data points
+        we are planning to have more discussion on how to define the continuation of a bottleneck **/
```
ian-r-rose, line 102: We do still want to do this in this PR. The part that we are deferring until later is figuring out the extent of congestion, not the 5-out-of-7 check

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
```diff
@@ -0,0 +1,48 @@
+{{ config(
+    materialized="incremental",
+    cluster_by=["sample_date"],
+    unique_key=["id", "sample_timestamp"],
+    snowflake_warehouse = get_snowflake_refresh_warehouse(small="XL")
```
ian-r-rose, line 5: Let's do small = "XS" and large = "XL"

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,117 @@
...
+
+bottleneck_checks as (
+    select
+        *,
+        case
+            when
+                speed_weighted < 40
+                and abs(distance_delta_ne) < 3
+                and speed_delta_ne <= -20
+                and speed_prev_ne >= 40
```
ian-r-rose, line 89: Do we need this check? Based on my read of the criteria, we don't care about the previous speed, just the speed at the current station. We do, however, care about the speed delta.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,117 @@
...
+                and (direction = 'N' or direction = 'E')
+                then 1
+            when
+                speed_weighted < 40
+                and abs(distance_delta_sw) < 3
+                and speed_delta_sw <= -20
+                and speed_prev_sw >= 40
```
ian-r-rose, line 96: Same question, I don't think this check is described in the criteria.

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
ian-r-rose, line 1: This model needs some YAML

transform/models/intermediate/performance/int_performance__bottlenecks.sql
ian-r-rose, line 1: This model also needs YAML

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,117 @@
...
+        /** this criteria below needs to remain true for five out of seven 5-minute data points
+        we are planning to have more discussion on how to define the continuation of a bottleneck **/
+        -- case
+        --     when
+        --         speed_weighted < 40
+        --         and abs(distance_delta) < 3
+        --         and (speed_prev_ne < 40 or speed_prev_sw < 40)
+        --         then 1
+        --     else 0
+        -- end as bottleneck_cont
```
ian-r-rose, line 110: Let's not keep around commented-out code

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,117 @@
...
+        /** to do: define the end of a bottleneck **/
+
+    from calcs
+    order by sample_timestamp asc, freeway asc, direction asc, type asc, absolute_postmile asc
```
ian-r-rose, line 114: No need to order these models, it conflicts with the `cluster_by` directive above and causes unnecessary work

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
```diff
@@ -0,0 +1,48 @@
+{{ config(
+    materialized="incremental",
+    cluster_by=["sample_date"],
+    unique_key=["id", "sample_timestamp"],
```
ian-r-rose, line 4: There may be some performance improvements if we also include `sample_date` in the unique keys, since that's the clustering key for the table

britt-allen: Makes sense!

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: Can you explain conceptually how it conflicts? I just want to deepen my understanding of what's happening. I'll remove though!

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: Does it now? 😮‍💨

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
britt-allen: 😔

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: Without this check we cannot distinguish if we are in the middle of a bottleneck or the start of one. This was something I discussed with @kengodleskidot and he seemed to agree. I believe that part of building this model is challenging the existing criteria and updating it logically if we see an opportunity to improve. Other than it not being in the existing criteria is there a logical reason why you would not include this to make the bottleneck_start criteria more sound?

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
britt-allen: Okay! This was copied and pasted code btw so this change likely needs to be corrected elsewhere too.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: Doesn't a bottleneck lasting for 35 minutes or more constitute it's extent? If not, can you point me to the block of text you are reading that makes you think otherwise?

transform/models/intermediate/performance/int_performance__bottlenecks.sql
ian-r-rose: When we cluster by `sample_date`, that sorts by that key. It's not a huge conflict, since sorting by `sample_timestamp` and `sample_date` are basically consistent, but it leaves us open to extra work.

Order by is most useful for final queries, but it is not super useful for intermediate results with Snowflake. Some more guidance is [here](https://docs.snowflake.com/en/sql-reference/sql/create-table#usage-notes)

transform/models/intermediate/performance/int_performance__bottlenecks.sql
ian-r-rose: Doesn't the `speed_delta` tell us whether we are in the start of a bottleneck or the middle of one? I don't really see what the `speed_prev` check adds on top of the `speed_delta` check. No objection to refining the criteria if we see room for improvement, to be clear.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: Thank you Ian!

transform/models/intermediate/performance/int_performance__bottlenecks.sql
ian-r-rose: I should be clearer: we've talked about both the spatial and the temporal extent of the bottlenecks. My understanding was we are deferring any treatment of the spatial extent of the bottleneck. The temporal extent, however(5 out of 7 measurements) is still useful for determining whether a bottleneck occurs for a sustained amount of time.

britt-allen: Understood and agree! Thanks for the clarity.

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
ian-r-rose: Yeah, we need to do a pass through the warehouse configs to clean them up. I'd like to take a closer look once some of the performance-related stuff is stable.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
ian-r-rose: One of the things that's not obvious about incremental models, is that when you add a clustering key, dbt itself adds an `order by` in the final sql (you can see that executing ins Snowflake when you run an incremental model). So it's possible to accidentally sort and then re-sort a table.

Probably not a big deal here since `sample_timestamp` is already consistent with `sample_date`, but in general it's good practice to avoid `order by` here unless we have a strong reason to do it.

ian-r-rose: This is something that @JamesSLogan has pointed out in the past, and may have some additional insights about ordering

JamesSLogan: No additional insights from me!

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
britt-allen: Should I change this on the bottlenecks model too or only on this one? @ian-r-rose

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: I agree!

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
ian-r-rose: Let's default to XS/XL, and adjust as needed

transform/models/intermediate/performance/int_performance__bottlenecks.sql
ian-r-rose: I'm a little agnostic, if you and @kengodleskidot think this is an important part of the check, I'd be okay with it, we should just document it as an additional criteria

mmmiah: This tag of starting bottleneck will be helpful when you will have continuous bottleneck for several stations. In that case you need to have separate tracking column. I did some QA/QC on outputs. It looks great to me. but It will be great if you can add the starting tracking of bottleneck. For example if your binary bottleneck results like this 1,1,1,1,1,0,1,1,.., then your first 1 (station will be tracked as 1 and then 7th number station will be also 1. This will help to identify particular location where congestion start and do some more engineering investigations on how to improve the traffic conditions.
image

transform/models/intermediate/performance/int_performance__bottlenecks.sql
mmmiah: The criteria can be dropped for bottleneck determination to make sure it is not impacting the determination of bottleneck, however, it can be useful to track which station is the beginning of bottleneck.

mmmiah (commented):
Overall great work@ britt. I think that you still did not consider 5, five minutes windows to determine the bottleneck. I will be back to review again Once you have more update. I have gone through the code and left some clarification comments. Thank you for doing all these.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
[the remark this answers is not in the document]
britt-allen: sw = south / west and ne = north / east. it's to account for the direction of highway when the postmile is sorted in one way which in our case is ascending

britt-allen:
thanks for doing QA/QC, the model and columns have changed since you last reviewed if you want to take another look @mmmiah

transform/models/intermediate/performance/int_performance__bottlenecks.sql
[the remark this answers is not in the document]
mmmiah: This makes sense

britt-allen:
These are current results of my QA/QC:

The green underlined data is highlighting the 5 instances of when the criteria is met out of a rolling window of 7

Screenshot 2024-06-28 at 1 04 50 PM

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
```diff
@@ -0,0 +1,48 @@
...
+    snowflake_warehouse = get_snowflake_refresh_warehouse(small="XS")
```
ian-r-rose, line 5: Can you put `big="XL" here?

transform/models/intermediate/performance/_performance.yml
```diff
@@ -343,3 +343,30 @@ models:
...
+      how long the bottleneck was active during that particular shift on that day. 2) Spatial Extent - For
+      each 5-minute period when the bottleneck was active we see how far upstream the bottleneck extends.
+      For this, at each time point, we find the farthest upstream detector with speed less than 40 mph and
+      use that location as the extent during that 5-minute period. To get a single extent for the entire
+      bottleneck, we take the median of the extents for each 5-minute period. 3) Delay - The delay is simply
+      the sum of the individual PeMS segment delays for the entire duration and spatial extent of the
+      bottleneck. The delay is calculated with respect to a threshold speed of 60 mph.
+    columns:
+      - name: tk
+        description: tk
```
ian-r-rose, line 372: Still TK?

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,120 @@
...
+calcs as (
+    select
+        *,
+        lag(speed_weighted)
+            over (partition by sample_timestamp, freeway, direction, type order by absolute_postmile asc)
+            as speed_prev_ne,
+
+        lead(speed_weighted)
+            over (partition by sample_timestamp, freeway, direction, type order by absolute_postmile asc)
+            as speed_prev_sw,
```
ian-r-rose, line 57: These two don't seem like they are actually used in any of the calculations. I think it should be safe to remove them.

mmmiah (approved):
Looks good to me!

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,120 @@
...
+
+        absolute_postmile
+        - lag(absolute_postmile)
+            over (partition by sample_timestamp, freeway, direction, type order by absolute_postmile asc)
+            as distance_delta_ne,
+
+        lead(absolute_postmile)
+            over (partition by sample_timestamp, freeway, direction, type order by absolute_postmile asc)
+        - absolute_postmile
+            as distance_delta_sw
```
ian-r-rose, line 76: Can you add a descriptive comment for future readers about how "upstream" and "downstream" is determined here? Something along the lines of:

> Absolute postmile increases going north and east. Therefore, when the direction of the freeway for a station is north or east, the "upstream" station has a smaller postmile, and we need to `lag` to get the speed there. When the direction is west or south, the "upstream" station has a larger postmile, and we need to `lead` to get the speed there.

@kengodleskidot does that description of how postmile works sound correct to you?

Based on what I wrote above, I think that the SW versions of the deltas might actually be reversed. Speed delta should always be the speed of the current station minus the speed of the upstream station. This might be able to be validated against a known bottleneck, if @kengodleskidot has any examples handy

britt-allen: Added, will make changes pending further review and feedback

britt-allen:
I excluded the following columns `exclude (latitude, longitude, bottleneck_check, bottleneck_check_summed)` so for anyone doing QA/QC you may want to temporarily add them back. QA/QC can also be done without them.

britt-allen:
Can I get final review and approval or feedback on this today so I can easily move on to other parts of the bottleneck problem? @ian-r-rose @kengodleskidot

ian-r-rose:
> Can I get final review and approval or feedback on this today so I can easily move on to other parts of the bottleneck problem? @ian-r-rose @kengodleskidot

I still think that my comment above might be something that needs correcting: a sign error in the speed delta computation will flag speed *increases* instead of *decreases* as bottlenecks. This is something that should be possible to check in QA for this PR.

ian-r-rose (approved):
Looks good to me, thanks @britt-allen!

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,118 @@
...
+        larger postmile, and we need to lead to get the speed there. */
+
+        speed_weighted - lag(speed_weighted)
+            over (partition by sample_timestamp, freeway, direction, type order by absolute_postmile asc)
+            as speed_delta_ne,
+
+        lead(speed_weighted)
+            over (partition by sample_timestamp, freeway, direction, type order by absolute_postmile asc)
+        - speed_weighted
+            as speed_delta_sw,
```
kengodleskidot, line 63: The speed_delta_sw looks off. Below is an example of a speed increase between the upstream (316438 @ 8.01 mph) and downstream (318612 @28.75 mph) stations but the calculation results in a negative value (-20.73 mph) which I believe we are assuming is a speed reduction based on the logic on lines 86 and 92. Traffic is travelling Westbound on I-80.
STATION_AGG_FIVE_MINUTES Model:
image
Bottleneck Model:
image

In the next example the speed_delta_ne looks ok. Below is an example of a speed increase between the upstream (3016071 @ 67.11 mph) and downstream (3001021 @ 67.35) stations and the calculation results in a positive value (0.25 mph) which I believe we are assuming is a speed increase based on the logic on lines 86 and 92. Traffic is travelling northbound on route 99.
STATION_AGG_FIVE_MINUTES Model:
image
Bottleneck Model:
image

@britt-allen let me know if you want to go over this.

kengodleskidot (changes requested):
@britt-allen this is coming along well but a few comments we can talk about or you can address directly.

transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
```diff
@@ -0,0 +1,48 @@
...
+
+real_detector_status as (
+    select
+        station_id as id,
+        status,
+        sample_date,
+        lane
+    from {{ ref ("int_diagnostics__real_detector_status") }}
+    where status = 'Good'
+),
```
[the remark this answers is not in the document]
britt-allen: Let's discuss in our weekly session tomorrow @kengodleskidot

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: Based on what you posted in #260 and my latest correction the postmiles are being handled correctly.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
britt-allen: let's discuss this tomorrow in our modeling session, I am having trouble replicating

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -0,0 +1,118 @@
...
+station_five_minute as (
+    select
+        id,
+        sample_date,
+        sample_timestamp,
+        speed_weighted
+    from {{ ref ("int_clearinghouse__station_agg_five_minutes") }}
+    where {{ make_model_incremental('sample_date') }}
+
+),
```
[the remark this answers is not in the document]
ian-r-rose: The reason why the "Good" filter is done in the `int_clearinghouse__station_agg_five_minutes` model is that it aggregates data from the detector level to the station level (since bottlenecks don't really make sense at the lane level). We wanted to only include data we felt was reliable in the aggregation up to "station", so filtered the detectors for "Good" status. But since the status is only available at the detector level, it had to be done a bit earlier.

In the future, this model (and all the performance models) are intended to be based on imputed data, where there should be less need to filter for "Good", at least based on my understanding of the plan. So I think this model `int_clearinghouse__station_agg_five_minutes` would be considered a temporary model until the imputation is ready. Does that make sense?

transform/models/intermediate/performance/int_performance__bottlenecks.sql
ian-r-rose: Thanks for the nice example @kengodleskidot, looking forward to digging in tomorrow! As my dissertation advisor used to say "you only need to make an even number of sign errors!"

transform/models/intermediate/performance/_performance.yml
```diff
@@ -343,3 +343,40 @@ models:
...
           the number of lane-mile-month on the freeway lost due to operating under congested
           conditions instead of under free-flow conditions.
+  - name: int_performance__bottlenecks
+    description: |
+      The PeMS system runs a bottleneck identification algorithm every day. The original algorithm was
+      presented in, "Systematic Identification of Freeway Bottlenecks," by Chen, C., Skabardonis, A.,
+      Varaiya, P. (2003). Transportation Research Board, 2004.
+
+      This model identifies bottlenecks at every station. A bottleneck is defined by the following
+      conditions - 1) There is a drop in speed of at least 20 mph between stations, not time. 2) The speed
```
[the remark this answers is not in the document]
britt-allen: Good catch!

transform/models/intermediate/performance/int_performance__bottlenecks.sql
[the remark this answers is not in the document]
kengodleskidot: Thank you for the explanation and it makes sense to me.

transform/models/intermediate/performance/int_performance__bottlenecks.sql
kengodleskidot: I reviewed the same data points using the updated code (post merge) and the signs look good for speed_delta_sw values.

ian-r-rose: woo hoo!

Issue #214, opened 2024-05-20
When should we join the VDS raw data with the metadata?


ian-r-rose:
We've gone back and forth on this a few times with @kengodleskidot in the context of some of our large aggregations on 30-second VDS data. Some of our downstream metrics depend on detector metadata, which lives in the `STATION_META` and related tables, and not on the raw tables. In order to compute these metrics, we need to join the two tables together on `STATION_ID`.

Normally, we'd do that join as late as possible, so that filtering and especially aggregation can occur, allowing for a cheaper join. For instance, if we join on the 5-minute data, that's 10x smaller of a join than the 30 second data, so we would want to do that if we don't need any of the metadata to do the 5-minute aggregation.

However, a quirk of our base tables is that they are delivered in wide format, with `FLOW_1`, `FLOW_2`, `FLOW_3` ...`FLOW_8`, etc. But not every station has eight lanes of traffic! In fact, the modal one has two or three. So when we do some of our large aggregations, we are doing them on more lanes than exist, wasting a lot of effort.

Right now we are doing the aggregations first, and joining with the metadata second. But we should experiment with joining at other stages! We'd be paying higher costs for the joins, but lower costs on the aggregations.

Example DAGs showing the late join with the metadata:

<img width="1434" alt="image" src="https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/5728311/202c9106-f8c2-40b4-881f-f82f0f81ccdd">

<img width="1434" alt="image" src="https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/5728311/45462f74-b8b5-4d20-aca7-5a474c6782f9">

ian-r-rose:
@JamesSLogan any updates here?

JamesSLogan:
@ian-r-rose I am currently working on reworking the downstream models to use the updated 5 minute-aggregated detector model. One thing that the new metadata gives us access to is `detector_id`, so there are also a few places where joins can make use of this (instead of station id and lane).

ian-r-rose:
I'd be interested to hear your thoughts where `detector_id` could be most helpful! Agree that in many cases it's likely preferable to station-lane.

ian-r-rose:
Ideally, this metadata join should be able to cut down the number of rows in the 5-minute table by a factor of 5 or so by removing non-existent detectors, so I look forward to all our downstream models becoming significantly snappier!

PR #221, opened 2024-06-04
Included district in five_mins_station_agg file

mmmiah:


kengodleskidot (approved):
Looks good @mmmiah! You just need to figure out the dbt Cloud error before merging which I believe may be triggered since you are changing the structure of an incremental table by adding a new column.

transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql
```diff
@@ -62,6 +62,15 @@ aggregated_speed as (
         sum(volume * speed) / nullifzero(sum(volume)) as speed_weighted
     from station_raw
     group by id, lane, sample_date, sample_timestamp_trunc
+),
+
+aggregated_speed_by_district as (
```
ian-r-rose, line 67: This is an expensive join that we don't need to do here (at least, not yet! See some discussion in #214). The `district` field is already available in the `station_raw` staging model: https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/465d8dcd68cb7b56dde34b8b36345a82cfc2555d/transform/models/staging/clearinghouse/stg_clearinghouse__station_raw.sql#L4

So really we just need to include it in the final select statement (perhaps with an `any_value` aggregate).

The test failure that @kengodleskidot points out is actually a relevant one! The join here actually winds up with duplicate values for the unique combination of values `id`, `lane`, `sample_timestamp`. This is because `station_meta` has more than one entry for every ID due to containing the history of the station metadata.

mmmiah: Thank you ken and Ian for the quick review and valuable feedback. I just incorporated the suggestions. Would you mind to review very quickly and help to merge this to the main. Thanks!

ian-r-rose (approved):
Looks good, thanks @mmmiah!

Would you be willing to try to kick off a job for backfilling this [here](https://california.getdbt.com/deploy/6/projects/10/jobs/20)?

Issue #222, opened 2024-06-05
Daily Timeframe Definitions - Peak/Off Peak and AM/Noon/PM Shifts 


kengodleskidot:
For some of the analysis done in PeMS there are certain timeframes that are defined below:

Typically used for Performance Metrics (VMT, VHT, TTI, Q, Delay, Productivity)
AM Peak 6:00 AM - 9:59:59 AM
Day Off Peak 10:00 am - 2:59:59 PM
PM Peak 3:00 PM - 6:59:59 PM
Night Off Peak 7:00 PM - 5:59:59 AM

Typically used for Bottlenecks and Congestion:
AM Shift 5:00 AM - 9:59:59 AM
Noon Shift 10:00 AM - 2:59:59 PM
PM Shift 3:00 PM - 8 PM

We will want to explore creating seed files that contain these timeframes in a way they can be used to aggregate performance metrics for reporting.

kengodleskidot:
Based on our sprint meeting discussion (6/6/24) I will add these timeframes as variables in the dbt_project.yml file

junlee-analytica:
Variables have been added to dbt and yml files.

PR #227, opened 2024-06-10
Timeframe Variables

kengodleskidot:
Created variables to capture on-peak/off-peak as well as AM/Noon/PM shift timeframes for reporting purposes. This addresses #222. I have held off from updating the detector_status model to avoid potential impacts to the new imputation model that was recently merged (#209).


ian-r-rose (approved):
This looks good to me @kengodleskidot, thanks! It looks like there is a failure in the imputation model, which is unrelated to this PR. I'm looking into it, but feel free to merge at your convenience

Issue #212, opened 2024-05-20
Can we optimize how we pivot to long format in the raw 30-second VDS data?


ian-r-rose:
In #124 we took a table that was wide-formatted and pivoted it to long-format. We did this for good reasons, but it's not clear we did it in the most performant way. Basically, we did a pivot at the very base of the models so that we'd only have to do it once:

https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/465d8dcd68cb7b56dde34b8b36345a82cfc2555d/transform/models/staging/clearinghouse/stg_clearinghouse__station_raw.sql#L8-L48

Now, I am of the opinion that subsequent operations on this longer-pivoted data should have basically the same performance as the wide-formatted data. This is because:

1. The data are the same size, just in a different shape
2. The pivot-longer operation is fully local to a row. Not only is it embarrassingly parallel, it is also row-wise parallel.

That said, my opinion doesn't necessarily mean anything to the Snowflake optimizer. It may be that it does a poor job with this case, and cannot recognize the fact that the data are basically the same and it's effectively a local projection. So let's investigate!

1. Does a large aggregation (e.g. `int_clearinghouse__five_minute_agg`) cost the same as the equivalent on the wide data?
2. If not, is it worth the additional complexity to do our pivot-longer *after* our large aggregations?

Issue #213, opened 2024-05-20
 Can our two large aggregations share any code?


ian-r-rose:
We have two large incremental aggregations:

1. [int_clearinghouse__five_minute_station_agg.sql](https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/main/transform/models/intermediate/clearinghouse/int_clearinghouse__five_minute_station_agg.sql)
2. [int_diagnostics__samples_per_station.sql](https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/main/transform/models/intermediate/diagnostics/int_diagnostics__samples_per_station.sql)

Both of these are based on the 30 second raw data, and both are fairly expensive to compute. Is it possible to have them share any code so that we are only aggregating the raw 30 second data once? For instance, can we do something clever with the 5-minute aggregation, then compute the daily aggregation from that?

ian-r-rose:
Yes, exactly, I'm wondering if any of the large aggregations can be combined into a single model (probably the `five_minute_station_agg` model), and then we can aggregate *that* model to get the daily counts and aggregates (as well as hourly, etc).

It may be that there are blockers to doing it, I haven't looked in great detail, but I'm hoping that @tnrahim can take a closer look.

PR #230, opened 2024-06-10
Investigations: unpivoting raw data and performance

ian-r-rose:
An investigation into #212 

I'm not happy with the performance of the upivoting operation (implemented in #124, solving #121). I *think* it should be a relatively cheap operation, since every "unpivot" is local to its row, and could be done on-the-fly in memory. But it seems that the Snowflake optimizer thinks otherwise! I took a bit of a deep dive into the performance characteristics of our five-minute-station-aggregate model, which has been the most expensive model in the project so far.

**Learning 1**: Unpivoting using the [`FLATTEN`](https://docs.snowflake.com/en/sql-reference/functions/flatten) function is shockingly slow. This is the approach taken in #124, because it allows you to more easily unpivot multiple columns at once.

**Learning 2**: Unpivoting using the [`UNPIVOT`](https://docs.snowflake.com/en/sql-reference/constructs/unpivot) function is similarly slow. Internally, it probably does something very similar to `flatten`. Snowflake does not document it, but duckdb [describes](https://duckdb.org/docs/sql/statements/unpivot.html#internals) something very similar, where unpivot is implemented as a series of `unnest` functions.

**Learning 3**: Despite my sense of wanting to unpivot as early in the model hierarchy as possible (so that _all_ downstream models see long data), it seems that it's worth it to unpivot *after* the five minute aggregation, simply because there is less data to sort at the end of the day.

This PR investigates two different unpivoting approaches, before and after aggregating the data, for a total of four combinations:

**Approach 1:** Unpivot using the `flatten` function, as is done in `main`.

**Approach 2:** *Manually unpivot by individually selecting the "wide" columns, naming them the same thing, and combining the subsets using `UNION ALL`.

The following are some timing data for the experiments in this PR. I ran everything on an XL warehouse, and numbers are in seconds:

|    | `FLATTEN` | `UNION ALL` |
|---|------------|--------------|
| Pre-agg |    33m 28s    |          3m 14s             |
| Post-agg | 15m 2s     |            91s       |

Clearly, it makes a huge difference! The approach of aggregating first, then using `UNION ALL` instead of `FLATTEN` cuts down the time by a factor of 20!

@tnrahim would you mind reviewing this to see if you agree with the analysis? If so, I think we should take this approach in our basal models. It will mean somewhat more complex/repetative code, but as you can see from this PR, we can cut down on the repetition using jinja.


kengodleskidot:
I wanted to follow up to this topic by confirming if 14 fields/lanes are needed for flow, occupancy and speed at this time. Based on my review of the raw data tables in the data lake the various district schemas have 10 fields/lanes for flow, occupancy and speed.

transform/models/intermediate/db96/int_db96__five_minute_station_agg_flatten.sql
mmmiah, line 1: is this lane number 14 is consistent over historical time period? If not, is there a way, we can determine n_lanes value based on availability of lane number like have ' _n' at the end of 'lane_'.

mmmiah, line 1: Can we generate this list of number 'lateral flatten([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]) as lane' based on highest lane number?

transform/models/intermediate/db96/int_db96__five_minute_station_agg_unpivot.sql
mmmiah, line 1: same questions here: is this lane number 14 is consistent over historical time period. If not, is there a way, we can determine n_lanes value based on availability of lane number like have _n at the end of lane_n.

mmmiah (commented):
Looks great! Left couple of comments!

ian-r-rose:
Any updates here @tnrahim?

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
```diff
@@ -2,10 +2,11 @@
     materialized="incremental",
     cluster_by=["sample_date"],
     unique_key=["id", "lane", "sample_timestamp"],
```
ian-r-rose, line 4: Let's also include `"sample_date"` among the unique keys to hint to the Snowflake optimizer that we can merge based on that criterion.

transform/models/intermediate/db96/int_db96__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,91 @@
+{{ config(
+    materialized="incremental",
+    cluster_by="sample_date",
+    unique_key=["id", "lane", "sample_timestamp"],
```
ian-r-rose, line 4: Same here, I think it may be helpful to the optimizer to include `"sample_date"` in the unique keys (even though it is redundant with the timestamp). I haven't done profiling on it, however, just a hunch that including the clustering key in merge criteria is helpful.

transform/models/intermediate/db96/int_db96__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,91 @@
...
+                and sample_date
+                >= dateadd(
+                    day,
+                    {{ var("dev_model_look_back") }},
+                    current_date()
+                )
+            {% endif %}
+    {% elif target.name != 'prd' %}
+        where sample_date >= dateadd(day, {{ var("dev_model_look_back") }}, current_date())
+    {% endif %}
```
ian-r-rose, line 46: Let's use the `{{ make_model_incremental("sample_date") }}` macro here.

transform/models/intermediate/diagnostics/int_diagnostics__samples_per_station.sql
```diff
@@ -9,7 +9,7 @@ with
 
 source as (
     select *
-    from {{ ref ('stg_clearinghouse__station_raw') }}
+    from {{ ref ('int_clearinghouse__detector_agg_five_minutes') }}
```
ian-r-rose, line 12: This is a major change to the meaning of this model, since this model involves many 30-second sample counts. This changes it to count 5-minute aggregate samples (effectively reducing by a factor of 10 the counts). Please revert this change and instead apply the unpivoting-with-jinja logic to this model.

transform/models/intermediate/performance/int_performance__detector_metrics_agg_five_minutes.sql
```diff
@@ -7,8 +7,18 @@
...
+        sample_date,
+        sample_timestamp,
+        lane,
+        district,
+        avg(occupancy) as occupancy_avg,
+        sum(volume) as volume_sum,
+        sum(volume * speed) / nullifzero(sum(volume)) as speed_weighted
+    from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
     where {{ make_model_incremental('sample_date') }}
+    group by id, lane, sample_date, sample_timestamp, district
```
ian-r-rose, line 21: This is another change to this downstream model, aggregating based on something that has already been aggregated. I'd like this PR to make changes to *only* the `int_clearinghouse__detector_agg_five_minutes` and `int_diagnostics__samples_per_station` models. To downstream models it should look 100% the same, we're just doing a performance optimization.

transform/models/intermediate/diagnostics/int_diagnostics__constant_occupancy.sql
```diff
@@ -16,6 +16,18 @@ source as (
...
+    select
+        id,
+        sample_date,
+        sample_timestamp,
+        lane,
+        district,
+        avg(occupancy) as occupancy_avg
+    from source
+    group by id, lane, sample_date, sample_timestamp, district
+),
```
ian-r-rose, line 29: This column is already averaged and grouped by these things (you did it in the `int_clearinghouse__detector_agg_five_minutes`). The issue here is that you changed the name of the occupancy column in the upstream model. For the most part, this PR should not need any changes to any downstream models based in the five-minute-aggregate model.

ian-r-rose (commented):
Making good progress, but ideally we'd like to see *no* changes to downstream models: this is a performance optimization for `int_clearinghouse__detector_agg_five_minute` and `int_diagnostics__samples_per_station` where the resulting tables should look identical. Any changes to models downstream of that should be indicative of something that needs correcting.

ian-r-rose:
@tnrahim I pushed a workaround here for the weird inconsistency in how sqlfluff was formatting `loop.last`. It seems that it's a bug others have run into, and is being tracked here: https://github.com/sqlfluff/sqlfluff/issues/5371

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
```diff
@@ -20,27 +21,65 @@ with station_raw as (
...
+    count_if(volume is not null and occupancy is not null)
+        as sample_ct,
+        -- Sum of all the flow values
+    sum(volume) as volume_sum,
+    -- Average of all the occupancy values
+    avg(occupancy) as occupancy_avg,
+    -- calculate_weighted_speed
+    sum(volume * speed) / nullifzero(sum(volume)) as speed_weighted
+from agg_unioned
+group by id, lane, sample_date, sample_timestamp, district
```
ian-r-rose, line 85: This is a redundant aggregation, the jinja-SQL above already does exactly these (except for the `sample_ct` column, which looks like it's been dropped).

Making this table look the same as the previous one is mostly a matter of naming the columns the same thing, not of doing extra aggregations.

transform/models/intermediate/diagnostics/int_diagnostics__samples_per_station.sql
ian-r-rose, line 1: Nice! Looks like you're trying to get #213 in here as well?

ian-r-rose (commented):
Other than including `"sample_date"` in the unique keys for the two five-minute-aggregates, this is looking quite close to me. Let's get CI passing and merge!

ian-r-rose (commented):
@tnrahim I notice that there is some inconsistency in the use of "flow" vs "volume". They are basically synonymous, but more of our downstream models are using "volume". Can you rename the "flow" columns to "volume" columns for consistency?

It also looks like this has caused a problem or two in CI: https://california.getdbt.com/deploy/6/projects/10/runs/3783/

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
tnrahim: Added sample date to unique keys

transform/models/intermediate/db96/int_db96__detector_agg_five_minutes.sql
tnrahim: Added sample date to unique keys

transform/models/intermediate/db96/int_db96__detector_agg_five_minutes.sql
tnrahim: Updated to use     where {{ make_model_incremental('sample_date') }}

transform/models/intermediate/diagnostics/int_diagnostics__constant_occupancy.sql
tnrahim: This change has been reversed. int_clearinghouse__detector_agg_five_minutes has been updated to flow into this model without issues.

transform/models/intermediate/diagnostics/int_diagnostics__samples_per_station.sql
tnrahim: yes! It made sense to as we have modified the stage to no longer have aggregated date.

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
tnrahim: I've updated this logic to go in the correct spot. Apologize for committing this, I was just putting the names these while thinking things through.

transform/models/intermediate/db96/int_db96__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,93 @@
...
+         buckets, it will be assigned to the one latter one due to
+         the floor() call.
+        */
+        dateadd(
+            'minute',
+            floor(minute(sample_timestamp) / 5) * 5,
+            trunc(sample_timestamp, 'hour')
+        ) as sample_timestamp_trunc
+    from {{ ref("stg_db96__vds30sec") }}
+    {% if is_incremental() %}
```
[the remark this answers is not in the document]
tnrahim: Updated if is_incremental block in db96 agg_five_minutes with the macro make_model_incremental.sql in

transform/models/intermediate/diagnostics/int_diagnostics__samples_per_station.sql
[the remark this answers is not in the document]
tnrahim: I've updated samples_per_station to samples per detector and resolved downstream conflicts.

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
```diff
@@ -20,27 +21,82 @@ with station_raw as (
...
+            {{ lane }} as lane,
+            volume_{{ lane }} as volume_sum,
+            zero_vol_ct_{{ lane }} as zero_vol_ct,
+            occupancy_{{ lane }} as occupancy_avg,
+            zero_occ_ct_{{ lane }} as zero_occ_ct,
+            zero_vol_pos_occ_ct_{{ lane }} as zero_vol_pos_occ_ct,
+            zero_occ_pos_vol_ct_{{ lane }} as zero_occ_pos_vol_ct,
+            high_volume_ct_{{ lane }} as high_volume_ct,
+            high_occupancy_ct_{{ lane }} as high_occupancy_ct,
+            speed_weighted_{{ lane }} as speed_weighted
```
ian-r-rose, line 88: I don't think these have been added to the YAML for the model.

ian-r-rose (commented):
Code looks good to me, but I still see an issue in the CI run: https://california.getdbt.com/deploy/6/projects/10/runs/3847/

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
tnrahim: I've added fields without descriptions to the yaml for now, I'm not sure why the CICD continues create this table incorrectly within TRANSFORM_DEV"."DBT_CLOUD_PR_15_230_CLEARINGHOUSE" though.

ian-r-rose: It's not being cloned incorrectly, this is expected behavior of incremental models (see the docs [here](https://docs.getdbt.com/docs/build/incremental-models#what-if-the-columns-of-my-incremental-model-change)). I've pushed a config tweak that should help with CI, though a full refresh will still be required for the new columns

ian-r-rose: CI is now passing

tnrahim: Thank you! Are there any other config changes needed to let DBT know a full refresh is necessary when merging? Or can we go ahead and merge the branch?

ian-r-rose: We'll need to run a full refresh manually, I can handle that.

I think my last ask is to populate the `foobar` descriptions in the model yaml

ian-r-rose (commented):
Looks good, thanks @tnrahim!

britt-allen:
I'd love a share out on this issue @tnrahim @ian-r-rose this friday during mid sprint check in

britt-allen:
or thursday rather, so the full caltrans team gets a show and tell of this completed issue too @tnrahim @ian-r-rose

ian-r-rose:
What about during our sprint check-in this afternoon?

britt-allen:
yes i just edited my comment lol, would love that

PR #236, opened 2024-06-13
Handle duplicates in metadata

### LINKED ISSUE #228: Fix nightly build failure
The local imputation model from #209 is failing for both our [nightly job](https://california.getdbt.com/deploy/6/projects/10/runs/3229) as well as in some CI jobs (cf #227 ).

ian-r-rose:
Fixes #228. Handles the case where duplicated lanes due to data entry error in the metadata can cause problems in downstream models that need station-lane combinations to be unique.

In our case, a single data entry error was breaking a downstream model which required lane-stationid-timestamp to be unique.


transform/models/intermediate/diagnostics/int_diagnostics__real_detector_status.sql
mmmiah, line 1: 'qualify row_number() over (partition by station_id, lane_number, meta_date order by detector_id) = 1' this line code make sense to drop the duplicate lanes number

mmmiah (approved):
looks good to me! your qualify row number function literally solve this issue!

PR #234, opened 2024-06-13
Regional regression

mmmiah:
This PR updated the codes to deploy regional and local regression together. The regional regression consider 5 miles upstream and downstream ring buffer along the ML and HOV while local regression considered just upstream and downstream stations.  The regression type is tracked by boolean (local reg=true, regional regression=false). This PR is similar to local regression!

PR #238, opened 2024-06-17
Fix identification of stations needing imputation

ian-r-rose:
This fixes an issue with the initial selection of the five-minute data where I made a mistake in clause ordering, causing the samples requiring imputation to be filtered too early, so that no samples were actually being imputed (whoops!).

Thanks @mmmiah for the discussion here, I think this will fix the issue for both #209 and #234.


ian-r-rose:
Note that in #234 @mmmiah made an alternative fix by just removing out the filter for status = "Good". I think that's also a perfectly valid fix! My motivation for filtering for "Good" stations at this point was to make the left join slightly smaller and more efficient (though I haven't done the profiling to know whether it makes a big difference in the overall query performance).

mmmiah (approved):
This looks good to me and did some QA. It seems that now it imputes the missing data that was failed in last time!

Issue #185, opened 2024-04-25
AADT (Average Annual Daily Traffic)


jkarpen:
## Computation of AADT for ITS Stations
Another quantity that we precompute offline from traffic flow data is the AADT, or **Average Annual Daily Traffic**. Conceptually, this is simply the total vehicle flow per day, averaged over a year-long period. We compute the AADT in 8 different ways, for every ITS station in the system and for every 12-month period which starts at the first of a calendar month. This section describes the various algorithms used to compute these quantities.

Conventional methods for calculating AADT values require that the input data meet certain completeness standards. For example, the AASHTO's 1992 "Guidelines for Traffic Data Programs" requires that a full day's worth of flow data be available for every day of the week and for every month of the year, with no imputed data allowed. However, much real-world sensor data does not meet such rigorous standards. Several research groups have developed variant procedures designed to accomodate typical patterns of missing data and minimize their distorting effect on the averages. PeMS computes AADT values using 8 of these methods, which are summarized below. More details can be found in the paper "Using Incomplete Archived ITS Data to Calculate Annual Average Traffic Statistics" by Shawn Turner and Eun Sung Park (Proceedings of the Transportation Research Board 87th Annual Meeting, Washington, D.C., 2008). The inputs to these methods are **the total station flow values (across all lanes) aggregated to the hourly or daily level, depending on the algorithm**. In every case, **no imputed data is allowed**; only those hours or days for which all data points are real observed values are included in the computations.

The 8 methods used to calculate AADT in PeMS, for each station and each 12-month period, are as follows:
| AADT Method  | Description                                  |
|--------------|----------------------------------------------|
| AADT_1       | Arithmetic Mean                              |
| AADT_2       | ASTM Standard 1442                           |
| AADT_3       | Conventional AASHTO Procedures               |
| AADT_4       | Provisional AASHTO Procedures                 |
| AADT_5       | Sum of 24 Annual Average Hourly Traffic Volumes |
| AADT_6       | Modified ASTM Standard                       |
| AADT_7       | Modified Conventional AASHTO                 |
| AADT_8       | Modified Provisional AASHTO                  |

* AADT_1: Arithmetic Mean
This is simply the average of all daily traffic flows over the entire year.
* AADT_2: ASTM Standard 1442
[...]

Issue #241, opened 2024-06-18
Establish better naming convention for aggregate models


ian-r-rose:
As we've been adding new models our organization, our model names have become a bit of a mess, without much consistency to them. This is not really a bad thing, we've been learning more about the needs of the data models, and deciding upon useful naming conventions is something that really comes out of having a strong idea of these needs.

I've noticed a few pain points/inconsistencies with our aggregate models in particular:

1. It's usually not possible to tell from the name whether the aggregate is at the station or detector (or controller!) level. Sometimes the name says `station`, but it's actually at the detector level. Sometimes it says neither.
2. We don't have much consistency about how to name different levels of temporal aggregation
3. We don't have a convention for how to name different levels of spatial aggregation

I'm opening this issue to discuss possible naming conventions going forward. Here is one proposal (but I'm interested in hearing other ideas!)

### Proposal

Overall name structure is:

`{model stage}_{category}__{device type}_agg_{temporal aggregation}_{spatial_aggregation}`

This is pretty verbose, but would capture most of the relevant information about the *intent* of an aggregate model from the name. Some description of the components:

**model stage**: This is the stage of a model, such as `stg`, `int`, or as we move towards "mart" models, could be something like `dim`, or `fct`, or nothing.

**category**: This is either the data source system (e.g., `clearinghouse` or `db96`) or the broad category of intention of the data model (like `diagnostics` or `imputation`). This usually would correspond the the name of the directory in which the model file sits.

**device type**: For VDS data this would usually be "controller", "station", or "detector", and indicates the "grain" of the model. For instance, most of our VDS tables are at the "detector" level because they include a lane column. But when the lanes are aggregated up to the station level, it would be "station". Note that this could also be considered a spatial aggregation, but in most cases I felt it was important enough information to have a dedicated part in the name.

**temporal aggregation**: One of `five_minute`, `hourly`, `daily`, `weekly`, `monthly`, or `yearly`.

**spatial aggregation**: Things like `district` or `freeway` or `county` or `city`.

If one part of the naming convention didn't make sense for a given model, we could drop it from the name. In particular, I'm thinking that with "mart" models, which are intended to be shared with downstream data consumers who might not know our conventions and project structure, we might drop model stage and category.

A couple of examples:

* `int_clearinghouse__five_minute_station_agg` --> `int_clearinghouse__detector_agg_five_minute`
[...]

PR #242, opened 2024-06-18
AADT and K_factor

mmmiah:
This PR calculated 8 different AADT at the station level that is described details in issue #185. It also calculates K-factors for each station. Please have your review and point out if there is any missing logic. Thanks!


transform/models/intermediate/clearinghouse/_clearinghouse.yml
```diff
@@ -474,3 +474,67 @@ models:
           The lost Productivity perfromance metric is
           the number of lane-mile-month on the freeway lost due to operating under congested
           conditions instead of under free-flow conditions.
+  - name: int_clearinghouse__annual_average_daily_traffic
```
ian-r-rose, line 477: Per our conventions discussed in #241, I think this might be classified as a "performance-related" model, and be named something like `int_performance__station_annual_average_daily_traffic`. I think it's complex enough that it doesn't really need to fit into the specific naming convention for temporal aggregates, however

transform/models/intermediate/clearinghouse/int_clearninghouse__station_k_factor.sql
ian-r-rose, line 1: Can you add YAML for this model?

Similar to the above, I think this might be more appropriate in the "performance" directory

transform/models/intermediate/clearinghouse/int_clearninghouse__station_k_factor.sql
```diff
@@ -0,0 +1,43 @@
...
+        id,
+        district,
+        type,
+        observation_month,
+        kfactor_month,
+        hourly_volume as k_factor
+    from
+        traffic_with_rank
+    where
+        traffic_rank = 30
```
ian-r-rose, line 40: In #242 this is described as 

>  Along with the 8 AADT values, PeMS computes a "k-factor" for each station and for each month. This is defined to be the 30th highest hourly traffic flow observed at that station among all hours of the preceding year.

Is that description wrong? This computes the k-factor for the preceding month, but the description says for the preceding year.

ian-r-rose, line 40: There's an opportunity to consolidate these last two CTEs by using a [`qualify`](https://docs.snowflake.com/en/sql-reference/constructs/qualify) clause in `traffic_with_rank`:

```sql
qualify RANK() over (
    partition by id, observation_month
    order by hourly_volume desc
) = 30
```

ian-r-rose (commented):
This is really great work @mmmiah! There *may* be a way to combine a few of the different metrics and reduce the join count, but to be honest that's probably not worth any savings at that level of aggregation. The SQL looks well-structured and readable to me, though I don't really understand the motivation behind the metrics as a non-traffic engineer.

transform/models/intermediate/clearinghouse/int_clearninghouse__station_k_factor.sql
mmmiah: Yes, I think that it will be preceding years only, not the month, A single value in a particular year for a station
image

mmmiah: @ian-r-rose ,Great suggestion! I just incorporated your suggestion. Thank you!

mmmiah: I have also included K-50 and K_100, but I think that we are not going to get the factor as we have 8 AADT, We will leave the option for users to choose the AADT to get the final k_factors. We will just report 30th, 50th and 100th highest hourly volume in a particular year. We will discuss this things with PeMS committee also.

transform/models/intermediate/clearinghouse/int_clearninghouse__station_k_factor.sql
mmmiah: I have combined k factor model with AADT model and also included ymal for this model under performance. Thanks for your review. Appreciate!

transform/models/intermediate/clearinghouse/_clearinghouse.yml
mmmiah: @ian-r-rose ,Yes, named as 'int_performance__station_annual_average_daily_traffic' We have also k value in this model. how about naming it as ''int_performance__station_annual_average_daily_traffic_with_k_value (not exactly factor as we did not divide it with AADT, we just reported 30th, 50th and 100 th highest hour volume). Let me know if you have something better in mind

mmmiah:
This PR read for final review and merge! Thank you!

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,352 @@
...
+        district,
+        direction,
+        freeway,
+        type,
+        sample_year,
+        count(sample_month) as sample_ct,
+        avg(madt) as aadt_2
+    from madt
+    group by id, city, county, district, freeway, direction, type, sample_year
+    having count(id) >= 12
```
ian-r-rose, line 69: Can you expand on why we need this `having` statement? Is 12 corresponding to the number of months? Why would we expect there to be more than 12, if the previous aggregate is per-month?

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,352 @@
...
+        county,
+        district,
+        freeway,
+        direction,
+        type,
+        sample_year,
+        avg(aadw) as aadt_3
+    from aadw
+    group by id, city, county, district, freeway, direction, type, sample_year
+    having count(id) = 7
```
ian-r-rose, line 103: How often do we get a station with a different number than 7? Does that happen if, e.g., a station is only operational for a few days out of the year, and doesn't have samples for every day of the week?

ian-r-rose (approved):
This looks great to me @mmmiah! My only suggestion is that it might be helpful to add a few comments on your `having` statements to make it clearer what they are filtering out. But I see that it's well documented in the YAML, so I view the suggestion as optional

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: Yes, none of madt should be missing in 12 months for this aadt version. It should not be more than 12. I just replaced it as '=12' and also included the notes as 
    -- all 12 months average traffic volume is required to calculate this metric
    --other way, there should be no missing monthly average daily traffic

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: I have done sample count for this, it seems that data is not really complete for many cases. That's why we have 2nd version of this aadt which is named as aadt_7. 
AADT_7: Modified Conventional AASHTO
This procedure is a modified version of AADT_3, in which 1 of the 12 MADW values may be missing in the AADW subcomputation, and/or 1 of the 7 AADW values may be missing in the final AADT_7 computation. 

I have included the clarification notes for having count(id)=7.

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
ZhenyuZhu-Caltrans, line 1: for aadt2: 
having count(id) >= 12 ----> having count(sample_month) >=12

for aadt3:
I did not see the requirement of all average daily traffic of 7 days of the week needs to be nun-null to calculate the AADT. 
delete having count(id) = 7

for aadt4:
I did not see the requirement of all average daily traffic of 7 days of the week needs to be nun-null to calculate the AADT. 
delete having count(id) = 7

for aadt7:
I do not see the code for 
in which 1 of the 12 MADW values may be missing in the AADW subcomputation,

for aadt8:
I do not see the code for 
in which 1 of the 12 MADW values may be missing in the AADW subcomputation,

transform/models/intermediate/imputation/_imputation.yml
```diff
@@ -1,77 +1,77 @@
...
+          tagging wheather the other id is local counters
+          or regional counters. If the other id is just upstream or downstream
+          then it is considered as local counters. On the other hand
+          except local counters if any counters are located within 5 miles
+          upstream and downstream, then it is considered as regional id.
+          This boolean function where True means local counters and false means
+          regional counters.
+  - name: int_imputation__detector_agg_five_minutes
+    description: |
+      dfdf
```
britt-allen, line 50: This is incomplete

transform/models/intermediate/performance/_performance.yml
```diff
@@ -343,3 +343,73 @@ models:
           The lost Productivity perfromance metric is
           the number of lane-mile-month on the freeway lost due to operating under congested
           conditions instead of under free-flow conditions.
+  - name: int_performance__station_aadt_with_K_value
+    description: |
+      Station level annual average daily traffic and k value
```
britt-allen, line 348: What is k value?

transform/models/intermediate/clearinghouse/_clearinghouse.yml
britt-allen, line 1: I would add periods to the ends of all sentences. Some have them and others do not so it's a bit inconsistent.

transform/models/intermediate/performance/_performance.yml
britt-allen, line 1: Same here. I would add periods to the ends of all sentences. Some have them and others do not so it's a bit inconsistent.

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,362 @@
...
+        id,
+        city,
+        county,
+        district,
+        freeway,
+        direction,
+        type,
+        avg(daily_volume) as aadt_1,
+        date_trunc('year', sample_date) as sample_year
+    from {{ ref('int_performance__station_metrics_agg_daily') }}
```
britt-allen, line 15: you select from this table 2 times, the best practice it to put this in a CTE selecting all or only the columns you need. for example, 

`with daily_station_metrics as (
     select * from {{ ref('int_performance__station_metrics_agg_daily') }}
)`

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,362 @@
...
+-- AADT_2: ASTM Standard 1442
+madw as (
+    select
+        id,
+        city,
+        county,
+        district,
+        freeway,
+        direction,
+        type,
```
britt-allen, line 28: You can remove the 7 lines of code that it takes to individually list these columns (id, city, county, district, freeway, direction, type,) and replace it with a *

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,362 @@
...
+        id,
+        district,
+        type,
+        extract(dow from sample_date) as day_of_week,
+        extract(hour from sample_hour) as hour_of_day,
+        date_trunc('month', sample_date) as sample_month,
+        date_trunc('year', sample_date) as sample_year,
+        -- calculate monthly average flow for each hour of the week
+        avg(hourly_volume) as mahw
+    from {{ ref('int_performance__station_metrics_agg_hourly') }}
```
britt-allen, line 121: you select from this table 3 times, the best practice it to put this in a CTE selecting all or only the columns you need. for example, 

`with hourly_station_metrics as (
     select * from {{ ref('int_performance__station_metrics_agg_hourly') }}
)`

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,362 @@
...
+    select
+        id,
+        city,
+        county,
+        district,
+        direction,
+        freeway,
+        type,
+        sample_month,
+        sample_year,
```
britt-allen, line 48: same here, you can remove the 9 lines of code that it takes to individually list these columns (id, city, county, district, freeway, direction, type, sample_month, sample_year,) and replace it with a *

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,362 @@
...
+aadt_2 as (
+    select
+        id,
+        city,
+        county,
+        district,
+        direction,
+        freeway,
+        type,
+        sample_year,
```
britt-allen, line 64: same here, you can select *

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,362 @@
...
+-- AADT_3: Conventional AASHTO Procedures
+aadw as (
+    select
+        id,
+        city,
+        county,
+        district,
+        freeway,
+        direction,
+        type,
```
britt-allen, line 83: same here, you can select *

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,362 @@
...
+aadt_3 as (
+    select
+        id,
+        city,
+        county,
+        district,
+        freeway,
+        direction,
+        type,
+        sample_year,
```
britt-allen, line 101: same here, you can select *

britt-allen (commented):
This is looking good so far! Right now the file is extremely long and hard to parse so I recommend selecting star where you can. I also recommend making use of reusable CTEs rather than directly selecting from the same table multiple times. Even if you select from the table once it's a best practice to place your source data in a CTE.

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: It pops as ambiguous error if I do not choose the specific column before aggregating

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: It pops as ambiguous error if I do not choose the specific column before aggregating

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: It pops as ambiguous error if I do not choose the specific column before aggregating

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: It pops as ambiguous error if I do not choose the specific column before aggregating

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: dropped it and called from very first read

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: read once the daily and hourly file and then call the necessary column that I needed

transform/models/intermediate/performance/_performance.yml
mmmiah: added period for all ymal description

transform/models/intermediate/clearinghouse/_clearinghouse.yml
mmmiah: added the period at the end of all sentences

transform/models/intermediate/performance/_performance.yml
mmmiah: K Value is defined at the end of this model ymal as 
      - name: K_30
        description: 30th highest hour volume in the precceding year
      - name: K_50
        description: 50th highest hour volume in the precceding year
      - name: K_100
        description: 100th highest hour volume in the precceding year

Also edited the model description as:   - name: int_performance__station_aadt_with_K_value
    description: |
      Station level annual average daily traffic and k th highest hour volume in the precceding year

transform/models/intermediate/imputation/_imputation.yml
mmmiah: Let me do that in imputation branch and keep this separate for now

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
mmmiah: Changed-
for aadt2:
having count(id) >= 12 ----> having count(sample_month) =12
for aadt3:
Very less clear description. I have read through other resources and found there is no requirements for this restriction
I have deleted 'having count(id) = 7'

For aadt_4:
 Dropped the condition 'having count(id) = 7'
for aadt7:
I thought 'or' indicates just to apply one condition. As it is said and/or, based on your suggestions, I have incorporated both condition as 
    -- 1 of the 12 MADW values may be missing in the AADW subcomputation.
    having count(madw) >= 11  
    -- 1 of the 7 aadw values may be missing for this final AADT calculation.
    having count(aadw) >= 6
for aadt8:
Applied both condition now as 
    -- 1 of the 12 MADW values may be missing in the AADW subcomputation
    having count(madw) >= 11
    -- 1 of the 7 aadw values may be missing for this final AADT calculation.
    having count(aadw) >= 6

mmmiah:
@ZhenyuZhu-Caltrans @britt-allen 
Can you have 2nd look for AADT and k factor model? I am planning to merge this in main before Wednesday sprint.

kengodleskidot (commented):
Looks great @mmmiah just some high-level comments that you can address before or after this PR.

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,407 @@
...
+
+monthly_average_days_of_the_week_traffic as (
+    select
+        id,
+        district,
+        type,
+        day_of_week,
+        sample_month,
+        sample_year,
+        sum(mahw) as madw
```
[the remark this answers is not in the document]
mmmiah: Applied additional filter to drop zero value of madw for aadt calculation. However, we will not filter out hourly zeros as at late night like 3 pm may be zero for the whole week that will not surprise me. However, you are right that monthly average day of week traffic zero does not make sense. I just applied additional filters to avoid zero skewed concern.

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,407 @@
...
+    select
+        id,
+        city,
+        county,
+        district,
+        freeway,
+        direction,
+        type,
+        -- calculate  Monthly Average Days of the Week (MADW).
+        avg(daily_volume) as madw,
```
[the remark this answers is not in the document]
mmmiah: The zero values of madw was not filtered previously that may potentially create null values. I have added additional filter for madw value in all CTEs

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,407 @@
...
+        direction,
+        freeway,
+        type,
+        sample_year,
+        count(sample_month) as sample_ct,
+        avg(madt) as aadt_2
+    from madt
+    group by id, city, county, district, freeway, direction, type, sample_year
+    -- all 12 months average traffic volume is required to calculate this metric
+    -- otherway,there should be no missing monthly average daily traffic
```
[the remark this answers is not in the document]
mmmiah: first line comments seems sufficient. Just dropped the 2nd line comment to avoid confusion

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
```diff
@@ -0,0 +1,407 @@
...
+aadt_1 as (
+    select
+        id,
+        city,
+        county,
+        district,
+        freeway,
+        direction,
+        type,
+        avg(daily_volume) as aadt_1,
```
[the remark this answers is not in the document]
mmmiah: If you see that some CTEs initially have only id, district, type, these CTEs specially depending on station hourly agg file where we do not have all details meta info due to expensive joining issues (previously discussed). We do not have those details in hourly model to have them here.

transform/models/intermediate/performance/_performance.yml
```diff
@@ -343,6 +343,76 @@ models:
...
+        description: |
+        Similarly, this procedure is a modified version of AADT_4, in which 1 of the 12 MADW
+        values may be missing in the AADW subcomputation, and/or 1 of the 7 AADW values may
+        be missing in the final AADT_8 computation. This is Modified Provisional AASHTO.
+      - name: K_30
+        description: 30th highest hour volume in the precceding year.
+      - name: K_50
+        description: 50th highest hour volume in the precceding year.
+      - name: K_100
+        description: 100th highest hour volume in the precceding year.
```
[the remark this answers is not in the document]
mmmiah: Yes, K-Values are based on observed data only, I think that I need to filter the data for good detectors only that is not implemented in current version. I believe that it is better to filter the good detectors in spatial temporal aggs as this model depend on previous aggregation. What do you think?

transform/models/intermediate/performance/_performance.yml
```diff
@@ -343,6 +343,76 @@ models:
           The lost Productivity perfromance metric is
           the number of lane-mile-month on the freeway lost due to operating under congested
           conditions instead of under free-flow conditions.
+  - name: int_performance__station_aadt_with_K_value
+    description: |
+      Station level annual average daily traffic and k th highest hour volume in the precceding year.
```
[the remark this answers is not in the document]
mmmiah: Included the additional resource link in the ymal file. Currently, there is no imputed data used for this AADT calculation as all calculation depends on source file 'int_performance__detector_metrics_agg_five_minutes'. We did not filter out bad detectors in this PR. How about filtering the good detectors in spatial temporal aggregation first, then our this model should be as it is?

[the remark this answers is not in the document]
mmmiah: @kengodleskidot based on our discussion, I have filtered out the good detectors first, then aggregated volume in station level to further calculate the AADT and k value!

transform/models/intermediate/performance/_performance.yml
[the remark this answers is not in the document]
mmmiah: Data filtered for good detectors only based on discussion

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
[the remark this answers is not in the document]
mmmiah: Some of the aadt (like aadt_1,6,7 and 8) still may be null due to not having enough sample to satisfy the condition such as 
having count(sample_month) = 12 for aadt_2
 having count(madt) >= 11 for aadt_6
having count(madw) >= 11 and having count(aadw) >= 6 for aadt_7,
having count(madw) >= 11 and having count(aadw) >= 6 for aadt_8

mmmiah:
@kengodleskidot , ready for your re-review!

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
britt-allen: can you share a screenshot of the error? i'm not familiar

mmmiah: After last edit and revision, right now under each CTE there is some calculation needed, therefore, I believe that we can not simply select everything using *.  If you select everything from the CTE and then join or aggregate sometimes it says ambiguous error. Will share such things when face again for that specific case

transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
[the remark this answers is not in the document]
kengodleskidot: Thank you @mmmiah! I agree that the hourly data can contain zero values and your example is a good one. My main concern is if we are seeing daily volume values of 0 along the ML and HV stations which I do not believe should be the case unless all lanes for a station are impacted by an incident, construction or some other external issue. Regardless I do not believe we would want to use a daily volume value of 0 in the AADT calculation.

kengodleskidot (approved):
Looks great @mmmiah! Something that may warrant additional discussion is how the AADT values are affected when some samples are missing, e.g. 110 samples received in an hour instead of 120 expected.

ian-r-rose:
Wooo! Congrats @mmmiah, and thanks for the detailed review @kengodleskidot!

PR #243, opened 2024-06-18
renamed all models in Spatial_temporal_agg

mmmiah:
Renamed the models based on naming proposal from @ian-r-rose


mmmiah:
@ian-r-rose , could you please have a quick review on this renaming and merge with main, so that I can update the upstream model like AADBT and k-factor one. Thank you!

ian-r-rose:
Thanks for the quick turnaround on this @mmmiah! Since these models have performance metrics in them like "delay" and "q", I think they might be better classified in the `intermediate/performance` directory (with corresponding names). I think of the "intermediate/clearinghouse" models as ones that are very simple and don't have much additional business logic in them (effectively a slight step beyond "staging" models).

Also, what do you think about my proposed amendment to the convention here? https://github.com/cagov/caldata-mdsa-caltrans-pems/issues/241#issuecomment-2179084366 . In that case, these names would then be, e.g. `int_clearinghouse__station_agg_daily` --> `int_performance__station_metrics_agg_daily`

As long as you're doing this, do you think you could also rename the following?

`int_clearinghouse__five_minute_station_agg` --> `int_clearinghouse__detector_agg_five_minutes`
`int_diagnostics__samples_per_station` --> `int_diagnostics__detector_counts_agg_daily`
`int_imputation__five_minute_aggregation` --> `int_imputation__detector_agg_five_minutes`
`int_imputation__five_minute_aggregation_regional_reg` (to be consolidated with above)
`int_performance__five_min_perform_metrics` --> `int_performance__detector_metrics_agg_five_minutes`

ian-r-rose (approved):
Thanks for taking this on @mmmiah!

PR #248, opened 2024-06-21
Constant occ fix

britt-allen:
Implemented a filter to ensure a 48-row window


britt-allen:
@ian-r-rose i am having trouble debugging the test case we identified because my dev data is a smaller set and it is not in there. any recs?

ian-r-rose:
> @ian-r-rose i am having trouble debugging the test case we identified because my dev data is a smaller set and it is not in there. any recs?

Now that the nightly build has run, you should be able to use deferral and point to the prod version of the five-minute data.

ian-r-rose:
For the constant occupancy model, you can temporarily adjust the time frame in the initial CTE to make it look at the specific date you are interested in

PR #249, opened 2024-06-21
implemented incremental model logic as a macro and applied it across all relevant models

### LINKED ISSUE #225: Repurpose incremental logic as a macro
- Could re-write as a macro
- Stretch goal could be to publish a dbt package

dbt [docs](https://docs.getdbt.com/docs/build/incremental-models)

britt-allen:
fixes #225

### DIFF
--- a/transform/macros/_macros.yml
+++ b/transform/macros/_macros.yml
@@ -13,9 +13,7 @@ macros:
   - name: make_model_incremental
     description: |
       This allows for the implementation of complex incremental model
-      logic in one line and one place. It checks for the incremental model
-      configuration at the top of the file then implements a conditional
-      check of the target environement. Check out dbt's docs on incremental
+      logic in one line and one place. Check out dbt's docs on incremental
       models: https://docs.getdbt.com/docs/build/incremental-models-overview.
       In addition to incrementality, it implements the variable
       "dev_model_look_back", which limits the amount of data selected in dev.
--- a/transform/macros/_macros.yml
+++ b/transform/macros/_macros.yml
@@ -12,8 +12,8 @@ macros:
       * the `--full-refresh` flag is set via the CLI.
   - name: make_model_incremental
     description: |
-      This allows for the implementation of complex incremental model 
-      logic in one line and one place. It checks for the incremental model 
+      This allows for the implementation of complex incremental model
+      logic in one line and one place. It checks for the incremental model
       configuration at the top of the file then implements a conditional
       check of the target environement. Check out dbt's docs on incremental
       models: https://docs.getdbt.com/docs/build/incremental-models-overview.
--- a/transform/macros/_macros.yml
+++ b/transform/macros/_macros.yml
@@ -17,3 +17,8 @@ macros:
       configuration at the top of the file then implements a conditional
       check of the target environement. Check out dbt's docs on incremental
       models: https://docs.getdbt.com/docs/build/incremental-models-overview.
+      In addition to incrementality, it implements the variable
+      "dev_model_look_back", which limits the amount of data selected in dev.
+      Two things to note: 1) Users implement there own date column; and
+      2) This macros does not include a "where" keyword, but does include
+      "and". Users are expected to use a "where" clause when implementing this.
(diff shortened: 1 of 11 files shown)


transform/models/intermediate/diagnostics/int_diagnostics__constant_occupancy.sql
```diff
@@ -8,26 +8,13 @@
...
-        {% if target.name != 'prd' %}
-            and sample_date
-            >= DATEADD('day', {{ var("dev_model_look_back") }}, CURRENT_DATE())
-        {% endif %}
+    select *
+    from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
+
+    {{ make_model_incremental(
+        'sample_date') }}
+    and TO_TIME(sample_timestamp) >= {{ var("day_start") }}
```
ian-r-rose, line 16: This will fail when `target=prd` and when `is_incremental() = false` because in that case there will be no `where` clause and it will just be a bare `and`.

There's probably a way to set up the macro so that it *does* work in both cases, probably with an additional parameter. For instance, we might add a new parameter like `additional_filter_clause` that gets tacked on to the end of the incremental filtering. Not sure about the best way right now, but seems solvable.

transform/models/intermediate/diagnostics/int_diagnostics__samples_per_station.sql
```diff
@@ -8,24 +8,13 @@
...
-        {% elif target.name != 'prd' %}
-            and sample_date
-            >= DATEADD('day', {{ var("dev_model_look_back") }}, CURRENT_DATE())
-        {% endif %}
+    select *
+    from {{ ref ('stg_clearinghouse__station_raw') }}
+
+    {{ make_model_incremental(
+        'sample_date') }}
+    and TO_TIME(sample_timestamp) >= {{ var("day_start") }}
```
ian-r-rose, line 16: Same here

transform/macros/make_model_incremental.sql
ian-r-rose, line 1: Can you add some documentation for this macro to the yaml?

ian-r-rose (commented):
Thanks @britt-allen, this is looking great! Just a couple of minor issues

transform/macros/make_model_incremental.sql
britt-allen: Sí

transform/models/intermediate/diagnostics/int_diagnostics__constant_occupancy.sql
britt-allen: Ahh thank you for pointing that out. I'll work on a fix.

transform/macros/_macros.yml
```diff
@@ -10,3 +10,10 @@ macros:
 
       * The target table doesn't exist (i.e., hasn't been built yet)
       * the `--full-refresh` flag is set via the CLI.
+  - name: make_model_incremental
+    description: |
+      This allows for the implementation of complex incremental model
+      logic in one line and one place. It checks for the incremental model
```
ian-r-rose, line 16: I'm not sure what you mean by "it checks for the incremental model configuration at the top of the file". Are you referring to `is_incremental()`? I don't think that's an accurate description of what that macro does.

ian-r-rose: Maybe also worth noting a few things about the behavior of this macro in the docs:

1. It does not include a `where` keyword, but does include `and`. The user is expected to bring their own `where`.
2. In addition to incrementality, it implements `dev_model_look_back`, which limits the amount of data selected in dev. This is distinct from anything to do with incrementality.

transform/macros/make_model_incremental.sql
```diff
@@ -0,0 +1,28 @@
+
+{% macro make_model_incremental(date_col, multiplier=1) -%}
```
ian-r-rose, line 2: I don't really remember why I did the multiplier in that one model, but I suspect that it's not really relevant anymore. Maybe it's not worth implementing the `multiplier` parameter after all, and we can just remove it.

ian-r-rose (commented):
Looking good! Let's get this in today

ian-r-rose (approved):
Looks good!

PR #250, opened 2024-06-21
Imputation global regression

ian-r-rose:
This contains initial work towards #170.

A few notes from implementation:

1. There are too many merges with `station_meta`! I'm feeling like we should prioritize #214 and join at the five-minute unimputed stage. Station metadata is used in too many downstream models (including ones at the five minute level, both here and in `performance`) to not be joined at that stage.
2. Once we can eliminate `int_diagnostics__real_detector_status` and move the merge with the metadata to upstream models, all of these imputation models will become significantly simpler.

@mmmiah do you want to collaborate on this PR?


ian-r-rose (commented):
Nice delta @britt-allen!

ian-r-rose (commented):
@mmmiah This is ready for a look!

transform/models/intermediate/imputation/int_imputation__global_coefficients.sql
```diff
@@ -0,0 +1,174 @@
...
+                '2023-02-03'::date,
+                '2023-05-03'::date,
+                '2023-08-03'::date,
+                '2023-11-03'::date,
+                '2024-02-03'::date,
+                '2024-05-03'::date
+            ]
+        )
+    )
+),
```
ian-r-rose, line 25: I hope to replace this with the logic @mmmiah is developing in #264

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
mmmiah, line 1: `        local_imputed.regression_date as local_regression_date,
        local_imputed.volume_local_regression,
        local_imputed.volume_local_regression,
        local_imputed.occupancy_local_regression,
        local_imputed.occupancy_local_regression,
        local_imputed.speed_local_regression,
        local_imputed.speed_local_regression,
        regional_imputed.regression_date as regional_regression_date,
        regional_imputed.volume_regional_regression,
        regional_imputed.volume_regional_regression,
        regional_imputed.occupancy_regional_regression,
        regional_imputed.occupancy_regional_regression,
        regional_imputed.speed_regional_regression
        regional_imputed.speed_regional_regression,
    from unimputed
        global_imputed.regression_date as global_regression_date,
        global_imputed.volume_global_regression,
        global_imputed.occupancy_global_regression,
        global_imputed.speed_global_regression
    from unimputed_with_config`

Could we just report 'regression_date' instead of three different regression date (local, regional and global) in the final CTE . I believe that all three regression date across the row are same. We can just report one as 'regression_date'

mmmiah (approved):
Overall, the imputation looks great to me! Just left a minor comment.

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
ian-r-rose: It won't in general be the same `regression_date` for all three versions. For instance, we can have a situation where there are no coefficients available for "local" regressions (because none of the local stations were operational at the same time as the active station) but we do have coefficients available for "regional" regressions. In that case, `local_regression_date` will be null, and `regional_regression_date` won't.

ian-r-rose: When we combine them into a single model that selects one of the imputation methods as the final value, then we could also coalesce these three dates into one "regression_date".

Issue #115, opened 2024-03-14
Load VDS metadata from db96


ian-r-rose:
Now that we are no longer trying to load the raw configuration files (cf #75 ) we are planning to get the data from db96->(actually, detector config log and other related meta table is in DWO.)

ian-r-rose:
@pingpingxiu-DOT-ca-gov any updates here?

pingpingxiu-DOT-ca-gov:
Daily Update: 
I fixed the all NULL bug and there is additional problems with individual columns' type casting. 
Six table's dozens' of columns need to be proofread. 

@ian-r-rose @ZhenyuZhu-Caltrans @jkarpen @jasonlally

ian-r-rose:
@pingpingxiu-DOT-ca-gov when you are satisfied with your scripts to load the metadata, can you start loading them to `RAW_PRD`?

pingpingxiu-DOT-ca-gov:
I verified the data on RAW_DEV looks good. 

The sizes of the config tables are: 
11157, 58297, 61853, 5613692, 24448, 338813
for 

CONTROLLER_CONFIG, CONTROLLER_CONFIG_LOG, DETECTOR_CONFIG, DETECTOR_CONFIG_LOG, STATION_CONFIG, STATION_CONFIG_LOG

pingpingxiu-DOT-ca-gov:
Hi @ian-r-rose 

I need the right credentials, I currently hit below error: 
```
Executing command SNOWSQL_PWD="【REDACTED】" REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt snowsql -a NGB13288 -u MWAA_SVC_USER_PRD -o insecure_mode=True -f /nfsdata/dataop/uploader/snowsql/snowsql_load_DETECTOR_CONFIG_LOG.sql -d RAW_PRD
250001 (08001): Failed to connect to DB: NGB13288.snowflakecomputing.com:443. Incorrect username or password was specified.
```

pingpingxiu-DOT-ca-gov:
Done, I have updated the script. I also uploaded the six configuration tables to Prod. Count verifications showed they are the same counts to the Dev environment. 
@ian-r-rose

pingpingxiu-DOT-ca-gov:
Summarizing today's conversation: 
* Six configuration tables has been verified by Ian on Snowflake
* Daily refresh will happen after backfilling. 

[...]

PR #257, opened 2024-06-27
db96 config staging and intermediate models

ian-r-rose:
This adds staging and intermediate models for the new config metadata loaded in #115 .

Some implementation notes:

1. Now that we have more than one source of VDS data (clearinghouse and db96), I've created a new "vds" schema for models that are agnostic as to which data source is being used. Eventually, we want to union the historical data with the ongoing data, and I view that as happening in the `vds` schema.
2. As in other places, the `int_vds__*_config` models implement an SCD-2 pattern with `_valid_from` and `_valid_to` for when each particular metadata row is active.
3. I've done some additional denormalization in `int_vds__detector_config` to also attach the station config to those rows. This is a pragmatic step because so many downstream metrics rely on having station-level metadata, even when working at the detector level. @JamesSLogan `int_vds__detector_config` is likely the model that you will want to join with the five-minute data for #214


transform/models/intermediate/vds/int_vds__controller_config.sql
```diff
@@ -0,0 +1,33 @@
...
+    select * from {{ ref('stg_db96__controller_config_log') }}
+),
+
+config as (
+    select * from {{ ref('stg_db96__controller_config') }}
+),
+
+config_log_with_validity as (
+    select
+        *,
```
JamesSLogan, line 13: Does it make sense to exclude `time_id` here for the sake of eliminating redundant columns? (also in int_vds_station_config.sql, at least)

JamesSLogan (commented):
The model documentation was very helpful 👌

transform/models/intermediate/vds/int_vds__controller_config.sql
ian-r-rose: Yeah, I agree, I *almost* did that this morning, but was lazy

ian-r-rose: Done

JamesSLogan (approved):
Looks good to me

transform/models/_sources.yml
```diff
@@ -155,3 +155,256 @@ sources:
             description: The name of the file in the clearinghouse where the metdata was obtained from.
           - name: CONTENT
             description: The XML content of the status metadata file.
+  - name: db96
+    database: "{{ env_var('DBT_RAW_DB', 'RAW_PRD') }}"
+    schema: db96
+    description: |
+      Vehicle Detector Station (VDS) (aka Station) related data obtained via data relay from
+      db96 within the Caltrans network
```
britt-allen, line 163: nit-pick - sentence missing a period.

britt-allen: I noticed a few other instances in this file too.

transform/models/staging/db96/stg_db96__controller_config.sql
britt-allen, line 1: this file doesn't follow the dbt best practice, but if this is how all of our staging files are formatted i am fine with that.

britt-allen (approved):
Great work! Thank you for the detailed docs and carefully constructed models, Ian!

Issue #240, opened 2024-06-18
Consolidate local and regional regression models


ian-r-rose:
The models for local and regional regression are extremely similar (and rightly so, since they do more-or-less the same thing, just drawing from different pools of stations for regression!). It is probably possible to combine them to one model with columns for both local and regional regressions.

@mmmiah would you be interested in looking into this?

mmmiah:
@ian-r-rose , sure!

mmmiah:
@ian-r-rose , I think that we also want to consolidate the global one, once you are done!

ian-r-rose:
Agreed, we should look at that one next!

PR #258, opened 2024-06-28
Consolidation of local and regional regression

mmmiah:
This PR consolidate the local and regional regression estimates together based on issue #240. The join is done in a incremental form to avoid large joining over historical time period.

### DIFF
--- a/transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes_local_regional.sql
+++ b/transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes_local_regional.sql
@@ -1,0 +1,2 @@
+{{ config(materialized='table') }}
+
--- a/transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes_local_regional.sql
+++ b/transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes_local_regional.sql
@@ -1,2 +1,19 @@
 {{ config(materialized='table') }}
 
+--join regional regression with local regression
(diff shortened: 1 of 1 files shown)


mmmiah:
@ian-r-rose , any ideas why git pre-commit check is not happening here? Thank you!

ian-r-rose:
> @ian-r-rose , any ideas why git pre-commit check is not happening here? Thank you!

Good question, I'm not sure! Let's see if it persists, it may just be a glitch in github

ian-r-rose (commented):
Thanks @mmmiah, I was hoping that the local and regional versions of the regressions were similar enough that they could actually be combined into _the same file_, where the final table just has columns for both local and regional regressions. So we would actually get one fewer model, as they are both combined into `int_imputation__detector_agg_five_minutes_local_regional`.

If it's doable, that could save us a significant amount of effort in not needing to do as many large-to-large joins. Does that make sense?

mmmiah:
> Thanks @mmmiah, I was hoping that the local and regional versions of the regressions were similar enough that they could actually be combined into _the same file_, where the final table just has columns for both local and regional regressions. So we would actually get one fewer model, as they are both combined into `int_imputation__detector_agg_five_minutes_local_regional`.
> 
> If it's doable, that could save us a significant amount of effort in not needing to do as many large-to-large joins. Does that make sense?

I think that it is doable, let me try. I will kill this PR then Once I am done with that. Making this PR as draft, no need to review anymore

mmmiah:
Did this consolidation within the local regression

PR #259, opened 2024-06-28
combined regional regression estimates with local regression

mmmiah:
This PR consolidate the local and regional regression estimates together based on issue https://github.com/cagov/caldata-mdsa-caltrans-pems/issues/240.


transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
```diff
@@ -150,7 +150,37 @@ agg_with_local_imputation as (
...
+        id,
+        lane,
+        regression_date,
+        sample_date,
+        sample_timestamp,
+        detector_is_good,
+        imp_occupancy as occupancy_regional_regression,
+        imp_volume as volume_regional_regression,
+        imp_speed as speed_regional_regression
+    from {{ ref('int_imputation__detector_agg_five_minutes_regional_reg') }}
```
ian-r-rose, line 167: I'm envisioning `int_imputation__detector_agg_five_minutes_regional_reg` going away all-together, and this model would do *both* of the regressions. They share so much code that there is likely some performance benefit to doing them together.

So, when we join with the coefficients here, we would join with all the regional coefficients, but then do two different versions of the `samples_requiring_imputation_with_neighbors` CTE: one where `other_station_is_local` is true, and another where it is false. Does that make sense?

mmmiah: I see the goal is killing regional regression model and develop both model in a single script. Sure, The giant model and its ymal is up for your review. Once, you will see this model is good, I will kill regional individual model one **'int_imputation__detector_agg_five_minutes_regional_reg',** we do need this one any more. Thanks!

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
```diff
@@ -99,58 +99,124 @@ samples_requiring_imputation_with_coeffs as (
     where coeffs.other_station_is_local = true
 ),
 
--- Read the neighbours that have volume, occupancy and speed data.
+-- join the coeeficent with missing volume,occupancy and speed dataframe for regional regression
+samples_requiring_imputation_with_regional_coeffs as (
```
ian-r-rose, line 103: Looking at the query profile, the `asof join` here is one of the most time-consuming parts of the query. I think it might be possible to combine this `asof join` with the one above. If we do the following:

1. Do just one `asof` join, without filtering on `coeffs.other_station_is_local`
2. Move the `coeffs.other_station_is_local` check to the `samples_requiring_imputation_with_local_neighbors` CTE

There may be some nice performance improvements.

That said, I'm happy with this PR where it stands, and we could investigate the possible performance improvement in a follow-up!

ian-r-rose (approved):
Thanks @mmmiah, this looks fantastic!

ian-r-rose:
I took a quick look, and I now think that combining the two asof joins is not a super high priority here, it might help at the 10% level, but I'm not super worried about that right now!

@mmmiah do you want to remove the old regression file and then we'll merge?

mmmiah:
@ian-r-rose,  Sure, I have killed the regional regression model. We can merge this one with main now! Thank you for doing some efficiency test.

Issue #239, opened 2024-06-17
Create incremental model without hard-coded dates for regression coefficients


ian-r-rose:
In #234 and #209 we created a table of regression coefficients for imputation. That table could use some improvements:

1. It should make use of `pems_clearinghouse_start_date` to determine when to start
2. It shouldn't have hard-coded dates
3. It should be incremental so that as time goes, we can continually update the coefficients with new values

@mmmiah since you've been thinking about how often to compute the regression coefficients, would you be interested in taking this one on?

mmmiah:
@ian-r-rose, I can try, but feel free to assign 2nd expert person with me in case I need help!

PR #264, opened 2024-07-02
Imputation Hard Code replacement

mmmiah:
This a test PR to move forward and replace imputation hard code and also make the model incremental

Related to #239


mmmiah:
@ian-r-rose, I have started working on this issue. However, realized to keep you update and collaborate on this to have smooth replacement of hard code. I have replaced the hard code part, now I am thinking of making this model increment as it is costly now in the entire time window phase. Our existing incremental model look back for 7 days only. I think that we need bigger windows for imputation at least a season out of four season or a year. Any thoughts that I can follow for incremental model integration for this imputation model application? Thanks!

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
```diff
@@ -9,19 +9,24 @@ better-thought-out logic for what dates to evaluate regression
...
-                '2023-11-03'::date,
-                '2024-02-03'::date,
-                '2024-05-03'::date
-            ]
-        )
-    )
+with recursive date_sequence as (
+    -- Define the initial date, which is the earliest 3rd of February, May, August, or November from your dataset
+    select min(sample_date) as base_date
+    from {{ ref('stg_clearinghouse__station_raw') }}
```
ian-r-rose, line 15: Not a huge deal, but I think we should mostly try to standardize on using the 5-minute data for downstream logic rather than the 30-second one. Both should have the same list of sample dates.

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
```diff
@@ -9,19 +9,24 @@ better-thought-out logic for what dates to evaluate regression
...
+    where extract(month from sample_date) in (2, 5, 8, 11)
+    union all
+    -- Add three months to the previous date
+    select dateadd(month, 3, base_date)
+    from date_sequence
+    -- Ensure we don't go beyond the maximum date in your dataset
+    where
+        base_date <= (select max(sample_date) from {{ ref('stg_clearinghouse__station_raw') }})
```
ian-r-rose, line 23: This didn't seem to work in my testing: I saw 2024-08-01 when I ran the CTE

ian-r-rose (commented):
Thanks @mmmiah, this is very cool! I never quite feel smart enough to write a recursive CTE.

I noticed  of potential issue when I ran this CTE in a worksheet: it was surprisingly slow. It took 13 seconds to execute, and scanned 25B rows of data! This seems excessive to me to generate a small list of dates. Perhaps there is another way to generate them which is faster to execute?

I wonder if using [`dbt_utils.date_spine()`](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#date_spine-source) to generate your dates might be a bit faster? It looks like it could make monthly dates, then we could filter it down to the specific months we are interested in.

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
mmmiah: opps. I just fixed it. Now it should work properly

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
mmmiah: agreed. I have replace the source file to five mins one. I am testing the [dbt_utils.date_spine()](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#date_spine-source) function to see if it can do much better job

ian-r-rose:
I'll take a look today!

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
```diff
@@ -51,6 +57,7 @@ good_detectors as (
 
 agg as (
     select * from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
+    where {{ make_model_incremental('sample_date') }}
```
ian-r-rose, line 60: I think this `make_model_incremental` macro won't work well for this model. It probably needs something a bit more specific.

What I'm envisioning is something like:

1. Get the list of regression dates from your CTE as you have done
2. Figure out which regression dates have not been computed yet in the current model (`{{ this }}`)
3. Use that list of regression dates for computing the coefficients.

I think something like this could do the trick:

```sql
regression_dates_to_evaluate as (
    select * from regression_dates
    {% if is_incremental() %}
        minus
        select distinct regression_date from {{ this }}
    {% endif %}
)
```
that could then be used in the join with `agg` below.

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
```diff
@@ -1,27 +1,33 @@
...
+    where
+        extract(day from date_day) = 3
+        and extract(month from date_day) in (2, 5, 8, 11)
+),
+
+-- Select distinct regression dates
+regression_dates as (
+    select distinct base_date as regression_date
+    from date_sequence
+    order by base_date
```
ian-r-rose, line 30: There's usually no need to use `order by` in a intermediate CTE. I think you could probably combine this CTE with the previous one.

ian-r-rose (commented):
This is looking great @mmmiah! I'm not sure exactly what's causing the empty results, but I suspect that using a different approach for incrementality could help.

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
```diff
@@ -1,27 +1,33 @@
...
+    {{ dbt_utils.date_spine(
+        datepart="day",
+        start_date="'1998-10-01'",
+        end_date="current_date()"
+    ) }}
+),
+
+-- Filter dates to get the desired date sequence
+date_sequence as (
+    select cast(date_day as date) as base_date
```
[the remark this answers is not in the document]
mmmiah: The above code actually provides the date in below format-
1998-10-01T00:00:00
1998-10-02T00:00:00
1998-10-03T00:00:00
1998-10-04T00:00:00
1998-10-05T00:00:00
1998-10-06T00:00:00
There fore we need to cast only date and exclude time. I have revised the code and calling only date but still we need cast function to exclude time.
Revised the code as -
`with date_spine as (
    select cast(date_day as date) as date_day
    from (
        {{ dbt_utils.date_spine(
            datepart="day",
            start_date="'1998-10-01'",
            end_date="current_date()"
        ) }}
    ) as spine
),

-- -- Filter dates to get the desired date sequence
regression_dates as (
    select *
    from date_spine
    where
        extract(day from date_day) = 3
        and extract(month from date_day) in (2, 5, 8, 11)
),`

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
mmmiah: I see what we are looking for. just killed this CTE

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
```diff
@@ -50,7 +59,12 @@ good_detectors as (
 ),
 
 agg as (
-    select * from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
+    select
+        afm.*,
+        rd.*
+    from {{ ref('int_clearinghouse__detector_agg_five_minutes') }} as afm
+    left join regression_dates_to_evaluate as rd
+        on afm.sample_date = rd.regression_date
```
ian-r-rose, line 67: This join is already happening below in the `detector_counts` CTE. Importantly, it's an inner join, so dates from `agg` which are not within the time windows for the regression are dropped.

So the thing to do here is to update the below join to use `regression_dates_to_evaluate`.

mmmiah: ooops, you are right, did not notice really in down CTE

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
```diff
@@ -1,27 +1,36 @@
 {{ config(
-        materialized="table",
-        snowflake_warehouse=get_snowflake_warehouse(size="XL"),
-    )
-}}
+    materialized="incremental",
+    unique_key=['id','sample_date','sample_timestamp', 'lane'],
```
ian-r-rose, line 3: These unique keys are needed to determine which rows need to be updated in the incremental logic: we need to ask the question: "which columns uniquely specify a single regression model?"

In this case, I think it would be `id`, `other_id`, `lane`, `other_lane`, and `regression_date`. That is to say, for each regression date we have at most one regression between a pair of detectors.

ian-r-rose: @mmmiah I also think that correcting the unique keys here will fix the `'DBT_INTERNAL_SOURCE.SAMPLE_DATE'` issue you are seeing in CI

ian-r-rose (commented):
Other than updating the unique keys, I think this is looking ready to go in!

transform/models/intermediate/imputation/int_imputation__local_regional_regression_coefficients.sql
mmmiah: the existing snowflake table actually raises the error, I have updated the key also, it is running now for all historical period as the first time, will request your review once the run is done and will do little QC on output. Thanks for kind co-operation! Really appreciate!

mmmiah:
@ian-r-rose, ready for your re-review

ian-r-rose (approved):
This looks great @mmmiah! Pending CI, but the code looks good from my perspective

PR #268, opened 2024-07-03
fixed criteria definition based on feedback

britt-allen:


kengodleskidot (approved):
Looks good @britt-allen!

PR #291, opened 2024-07-11
Move metadata join upstream

### LINKED ISSUE #214: When should we join the VDS raw data with the metadata?

JamesSLogan:
Changes:
- Added metadata to 5 minute aggregation models (this gives us `detector_id`)
- Removed downstream metadata joins
- Disambiguated `id` by replacing with `station_id` 
- Updated some models to key off of `detector_id` instead of `station_id` and `lane`

Fixes #214


transform/models/intermediate/clearinghouse/int_clearinghouse__station_agg_five_minutes.sql
```diff
@@ -27,22 +27,28 @@ joined_data as (
...
         sample_date,
         sample_timestamp,
-        district,
+        any_value(freeway) as freeway,
+        any_value(direction) as direction,
+        any_value(station_type) as station_type,
+        any_value(absolute_postmile) as absolute_postmile,
+        any_value(district) as district,
+        any_value(latitude) as latitude,
+        any_value(longitude) as longitude,
```
ian-r-rose, line 45: Love this

transform/models/intermediate/performance/int_performance__station_metrics_agg_monthly.sql
```diff
@@ -52,7 +52,7 @@ monthly_station_level_spatial_temporal_metrics as (
 
         {% endfor %}
     from station_daily_data
-    group by id, sample_month, city, county, district, type, freeway, direction, length
+    group by station_id, sample_month, length
```
ian-r-rose, line 55: Here and above: is there a reason you haven't moved `length` into the `any_value` block?

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
```diff
@@ -1,13 +1,20 @@
 {{ config(
     materialized="incremental",
     cluster_by=["sample_date"],
-    unique_key=["id", "lane", "sample_timestamp"],
+    unique_key=["detector_id", "sample_timestamp"],
     snowflake_warehouse = get_snowflake_refresh_warehouse(small="XL")
```
ian-r-rose, line 5: I'm curious if you have done any testing on how much longer this table takes to build. Can we get away with bringing the warehouse sizes down?

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
```diff
@@ -13,12 +13,6 @@ with base as (
     where {{ make_model_incremental('sample_date') }}
 ),
 
-/* Configuration data for detectors, filtered to mainline or HOV, as the main travel lanes */
-detector_config as (
-    select * from {{ ref("int_vds__detector_config") }}
-    where station_type in ('ML', 'HV') -- TODO: make a variable for "travel station types"
-),
-
```
ian-r-rose, line 21: 🎉

ian-r-rose (commented):
Thanks @JamesSLogan this is really great work. I see there are a couple of small merge conflicts, and we may have to go through a couple of contortions to get CI passing with the incremental models, but overall I think this is looking basically ready to go in.

I'm curious if you did any performance testing to see how expensive this join is at the end of the day?

transform/models/intermediate/performance/int_performance__station_metrics_agg_monthly.sql
JamesSLogan: length is used in the delay calculations, as it turns out. It's probably possible to rework things to get around this but I largely stayed away from major logic changes.

ian-r-rose: I'm not sure I understand -- you've moved the other group-by keys into the list of `any_value`s, how is length different, regardless of how it's used downstream?

JamesSLogan: Fixed. Using `any_value(length) as length ` previously gave an error, but modifying the delay calculation to use the aggregated length fixes this.

ian-r-rose (approved):
Nice work @JamesSLogan!

PR #293, opened 2024-07-12
Fix asof join condition to include lane

ian-r-rose:
Came across this bug when working on some imputation cleanup. Without this we were matching too many coefficients and getting incorrect imputation values!


transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
mmmiah, line 1: At the beginning, we started lane to lane join, then we removed this lane to lane join and kept only id to id as we are getting too limited matching.  However, as we introduce three different regression, I believe that now we can be specific to lane to lane for better capture the traffic pattern along the same lane that you suggested. Good catch!

mmmiah (approved):

Issue #271, opened 2024-07-05
Create a final combined imputation model


ian-r-rose:
For the local, regional, and global imputation methods we have been storing the results in separate columns. This allows us to compare the results of different imputation methods.

That said, we also want to select one of the methods for each timestamp to use as our preferred value in downstream metrics calculations. Roughly speaking, we want the following prioritization:

1. Device-reported values if we consider them reliable
2. Local imputation
3. Regional imputation
4. Global imputation

Our final model should just have a single column for each of volume, occupancy, and speed. In addition, we want the following columns:

1. `observed`, boolean, indicating whether the value is observed or imputed
2. `imputation_method`, indicating which (if any) imputation method is used
3. `regression_date`, indicating the date that the regression coefficients were computed (if applicable)

At the end of the day, this model should look almost identical to `int_clearinghouse__detector_agg_five_minutes`, but with the additional columns indicating imputation info. This will make it easier to swap this new model into all of our downstream performance metrics.

@mmmiah are you interested in taking this one on?

mmmiah:
@ian-r-rose , yes, I can work on this.

kengodleskidot:
What are your thoughts on having the device reported values (observed) as it's own model and updating the `int_clearinghouse__detector_agg_five_minutes` model to contain the same values (plus the three new columns mentioned above) it currently has with the only difference being that the flow, occupancy, speed and sample count values would be populated with the results of the observed and imputed models based on the hierarchy in the initial comment? We can have a quick meeting to discuss (pros/cons) this if you'd like @ian-r-rose and @mmmiah.

ian-r-rose:
I'm not sure I understand your proposal @kengodleskidot, wouldn't that be mostly the same thing as creating a new model for final imputed values, and keeping `int_clearinghouse__detector_agg_five_minutes` as the device-reported ones?

kengodleskidot:
@ian-r-rose Yes, except that we would not need to change the other downstream model references to a new model since `int_clearinghouse__detector_agg_five_minutes` would have the "final" observed/imputed value based on the best data available. Creating a new model for the "final" values also works but just requires changing references in downstream models.

ian-r-rose:
> @ian-r-rose Yes, except that we would not need to change the other downstream model references to a new model since `int_clearinghouse__detector_agg_five_minutes` would have the "final" observed/imputed value based on the best data available. Creating a new model for the "final" values also works but just requires changing references in downstream models.

[...]

PR #297, opened 2024-07-12
Merging observed and imputated data

mmmiah:
This PR implemented the proposed #271 issue. The imputed and observed volume, speed and occupancy merged together to fill the hole in the data while tagged whether the value is imputed or observed and also imputation method and date was used to fill the gap.


mmmiah:
@ian-r-rose @kengodleskidot, this PR ready for your feedback! Thanks!

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,130 @@
...
+    snowflake_warehouse=get_snowflake_warehouse(size="XL")
+) }}
+
+-- read the observed five minutes data
+with
+observed_five_minutes_agg as (
+    select *
+    from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
+    where {{ make_model_incremental('sample_date') }}
+),
```
ian-r-rose, line 13: I think this data is already in `int_imputation__detector_agg_five_minutes`, so we should be able to do this without a big join between the two tables. If the data are *not* in `int_imputation__detector_agg_five_minutes`, it's probably better to add it there and use that in this model.

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,130 @@
+{{ config(
+    materialized="incremental",
+    unique_key=['id','lane','direction','sample_date', 'sample_timestamp'],
+    snowflake_warehouse=get_snowflake_warehouse(size="XL")
```
ian-r-rose, line 4: Let's use the `get_snowflake_refresh_warehouse` macro here, which uses a bigger warehouse whenever we do a full refresh

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,130 @@
...
+            volume_local_regression,
+            volume_regional_regression,
+            volume_global_regression,
+            speed_local_regression,
+            speed_regional_regression,
+            speed_global_regression,
+            occupancy_local_regression,
+            occupancy_regional_regression,
+            occupancy_global_regression
+        ),
```
ian-r-rose, line 71: With an exclude list this long, I'd probably instead just explicitly list the columns you *do* want

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,130 @@
...
+        coalesce(volume_sum, volume_local_regression, volume_regional_regression, volume_global_regression)
```
ian-r-rose, line 72: I think this does not correctly handle the case where `volume_sum` is non-null, but we don't consider it accurate (e.g., high-occupancy, or constant-occupancy diagnostics). In the imputation code, we currently only impute for values that we don't consider "Good", so anything that has a value for one of the imputation methods is already something that is not "Good".

I wonder if we should just include the diagnostic status in a column in the imputation table so that we could use it here. It would make for a bigger table, but probably not prohibitively so.

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,130 @@
...
+            as volume_sum,
+        case
+            when station_type in ('ML', 'HV') then not coalesce(volume_sum is not null, false)
+        end as volume_sum_imputed,
+        case
+            when volume_sum is not null then 'detector_reported'
```
ian-r-rose, line 78: For consistency's sake, I think "observed" would be a better name than "detector_reported", unless there is a distinction I am missing.

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,130 @@
...
+            speed_regional_regression,
+            speed_global_regression,
+            occupancy_local_regression,
+            occupancy_regional_regression,
+            occupancy_global_regression
+        ),
+        coalesce(volume_sum, volume_local_regression, volume_regional_regression, volume_global_regression)
+            as volume_sum,
+        case
+            when station_type in ('ML', 'HV') then not coalesce(volume_sum is not null, false)
```
ian-r-rose, line 75: I don't think I understand this logic. Can you expand a bit? If we use the observed value, then I think this evaluates to `false`. If we use an imputed value, then this also evaluates to `false`. If `volume_sum` is null (which will happen if all of the imputation methods fail), this evaluates to `true`.

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,130 @@
...
+            when volume_regional_regression is not null then 'regional'
+            when volume_global_regression is not null then 'global'
+            else 'none'
+        end as volume_imputation_method,
+        case
+            when volume_sum is not null then null
+            when volume_local_regression is not null then local_regression_date
+            when volume_regional_regression is not null then regional_regression_date
+            when volume_global_regression is not null then global_regression_date
+        end as volume_imputation_date,
```
ian-r-rose, line 89: For clarity, maybe we should name this `regression_date`. Right now, the logic for `volume_regression_date`, `occupancy_regression_date`, and `speed_regression_date` are identical. Any reason not to combine them into a single field?

ian-r-rose (commented):
Thanks @mmmiah, this is a great start

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: Yes, I can remember we joined imputed and observed during the imputation phase. I will drop it

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: updated!

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: dropped exclude function and explicitly listed the items

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: diagnostic status column is already in our imputation table. I have updated the code based on detector status

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: updated as observed. Thank you!

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: updated the code with detector status and used new logic to report the imputation status. please review the new code.

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: agreed and update to just regression date

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,113 @@
...
+        freeway,
+        detector_is_good,
+        sample_date,
+        sample_timestamp,
+        speed_five_mins,
+        -- select the imputed value
+        case
+            when detector_is_good = false
+                then
+                    coalesce(volume_sum, volume_local_regression, volume_regional_regression, volume_global_regression)
```
ian-r-rose, line 35: Here and below: this is still wrong for the case where `detector_is_good = false`, but `volume_sum` is non-null. In that case, we will report the observed (bad) volume, rather than our imputed values. What we want here is a coalesce statement that *does not include* the observed column

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
```diff
@@ -0,0 +1,113 @@
...
+        end as occupancy_imputation_method
+    from obs_imputed_five_minutes_agg
+),
+
+final_imputation_with_tag as (
+    select
+        *,
+        coalesce(detector_is_good = false and volume_sum is not null, false) as is_volume_imputed,
+        coalesce(detector_is_good = false and speed_weighted is not null, false) as is_speed_imputed,
+        coalesce(detector_is_good = false and occupancy_avg is not null, false) as is_occupancy_imputed
```
ian-r-rose, line 108: I don't think these need to be coalesce statements, both conditions are boolean and should never result in `null`. It can just be `detector_is_good = false and `volume_imputation_method is not null as is_volume_imputed`. Or, to be honest, I think we could skip this tag all together, and people can check the `volume_imputation_method` to see whether it's an imputed value.

ian-r-rose (commented):
Thanks @mmmiah, this is looking close

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
ian-r-rose, line 1: Can you add some YAML documentation for this model?

transform/models/intermediate/imputation/int_observed_imputated__detector_agg_five_minutes.sql
mmmiah: right point. I am going drop this tagging

ian-r-rose:
Can you yaml docs here? Since this is going to be one of our cornerstone models, I think it's especially important that this one has good docs.

transform/models/intermediate/imputation/_imputation.yml
```diff
@@ -77,7 +80,103 @@ models:
...
           were evaluated.
+      - name: regional_regression_date
+        description: |
+          The date on which the coefficients in the regional regression
+          were evaluated.
+      - name: global_regression_date
+        description: |
+          The date on which the coefficients in the global regression
+          were evaluated.
+  - name: int_observed_imputated__detector_agg_five_minutes
```
ian-r-rose, line 101: Can we name this file `int_imputation__detector_imputed_agg_five_minutes`?

ian-r-rose (approved):
Looks like some minor formatting stuff in CI, but the code looks great to me, thanks @mmmiah!

transform/models/intermediate/imputation/_imputation.yml
mmmiah: sure! Done!

mmmiah:
Closed with notes as completed!

ian-r-rose:
Woohoo!

Issue #262, opened 2024-07-02
Create dbt post-hook or similar for unloading marts


jkarpen:
For final mart tables we want to unload them to S3 for use by dashboarding tools. Snowflake supports unloading to external stages. A dbt table materialization won't do that by default so we need to create a post-hook (there may be other strategies instead) to unload marts to public S3 bucket. 

May start with incomplete subset of the data early before switching to the "real" data later.

PR #303, opened 2024-07-16
Marts S3 bucket

### LINKED ISSUE #253: Create S3 stage for unloading marts
Once we have tables that we want to distribute to a future website or open data, we'll want to unload them to S3 (or similar) for consumption. This S3 bucket can be open. I'm not sure the best way to operationalize this right now, but a few options are:

1. Using a dbt post-hook to unload tables after being built
2. Using a custom materialization for unloading to S3
3. Using iceberg tables in an S3 bucket maintained by us

ian-r-rose:
This creates two public S3 buckets (one for dev and one for prod) for unloading data marts. It also creates associated stages in Snowflake which allow the `TRANSFORMER_{env}` roles to unload data using a `COPY INTO` command.

Fixes #253


ian-r-rose:
I've also gone ahead and implemented #262 here. By default, *any* table which is built in the `marts` directory will automatically be unloaded to S3 as part of the nightly build. You can optionally specify a partitioning for larger tables, but for truly large tables (like the 5-minute aggregates), we probably want a different strategy that uses more incremental logic. I'd defer on that for now.

But for small to moderate sized tables (the kind you might want to point PowerBI to), this should basically work for anything in the `marts` directory. I've added a couple of examples in this PR, one of which uses partitioning, the other of which does not.

britt-allen (approved):
Looks good to me, thank you Ian!

PR #304, opened 2024-07-17
Adding duplicate_proof capability

pingpingxiu-DOT-ca-gov:
Add staging VDS30SEC's deduplication (with incremental)

### DIFF
--- a/transform/models/staging/db96/stg_db96__vds30sec.sql
+++ b/transform/models/staging/db96/stg_db96__vds30sec.sql
@@ -52,5 +52,5 @@ select
     speed_13,
     speed_14
 from {{ source('db96', 'vds30sec') }}
-where true and {{ make_model_incremental('sample_date') }}
+where {{ make_model_incremental('sample_date') }}
 qualify row_number() over (partition by vds_id, sample_date, sample_time order by vds_id) = 1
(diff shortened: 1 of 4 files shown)


pingpingxiu-DOT-ca-gov:
@ian-r-rose  could you give an approve of this?

transform/models/staging/db96/stg_db96__vds30sec.sql
```diff
@@ -46,3 +52,5 @@ select
     speed_13,
     speed_14
 from {{ source('db96', 'vds30sec') }}
+where true and {{ make_model_incremental('sample_date') }}
```
ian-r-rose, line 55: Do we need the `true` condition? I was under the impression that `where {{ make_model_incremental('sample_date') }}` would be enough.

Note that the `make_model_incremental` macro handles the case where there is no other condition already:
https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/78af5cbeffa5158a1302ccd41a7d052faf0fc968/transform/macros/make_model_incremental.sql#L27-L29

ian-r-rose (commented):
Looking good to me @pingpingxiu-DOT-ca-gov, just one question

transform/models/staging/db96/stg_db96__vds30sec.sql
pingpingxiu-DOT-ca-gov: Thanks @ian-r-rose . indeed the `true` part was not needed. I'm changing it

kengodleskidot (approved):
No additional comments on my end @pingpingxiu-DOT-ca-gov

ian-r-rose (approved):
Looks good to me, thanks @pingpingxiu-DOT-ca-gov. I'll merge and do a full refresh of this model today

PR #306, opened 2024-07-17
Bottleneck extent (spatial)

### LINKED ISSUE #255: Bottleneck Spatial
Bottleneck (Spatial) Extent: For each 5-minute period when the bottleneck was active we see how far upstream the bottleneck extends. For this, at each time point, we find the farthest upstream detector with speed less than 40 mph and use that location as the extent during that 5-minute period. To get a single extent for the entire bottleneck, we take the median of the extents for each 5-minute period.

britt-allen:
fixes #255


kengodleskidot (approved):
Great job @britt-allen, looks good to me!

britt-allen:
@kengodleskidot there's been a lot of changes to the code/logic since your last approval. can you review and QA/QC once more and provide final approval or feedback, please?

britt-allen:
For anyone reviewing I'd love to get your feedback and understanding of if this model is correctly identifying instances of bottleneck spatial extent. If not, please feel free to provide feedback in the form of screenshots, proposed code, etc. to validate your assumptions.

mmmiah (approved):
This is one of the great piece of works. Looks good to me @britt-allen

thehanggit (commented):
@britt-allen Thanks for your extraordinary work! I have spent time trying to understand the codes thoroughly. Despite the logic of defining bottlenecks (whether from upstream or downstream), one thing I noticed is that when calculating congestion_length, it also calculates the total length of uncongested_sequence, even though the length is defined as 0. Could we avoid that calculation because for most of time period, most contiguous stations are uncongested.  I suppose that would consume notable time to calculate the sum function for them (please correct me if I'm wrong). 
I will continue checking the codes on Tuesday, especially on the bottleneck logic compared with the paper and making necessary comments.

britt-allen:
Thank you for your investigation and feedback Hang! do you have time to go
through this together today at 12 PM?

sent from my iPhone

britt-allen:
> @britt-allen Thanks for your extraordinary work! I have spent time trying to understand the codes thoroughly. Despite the logic of defining bottlenecks (whether from upstream or downstream), one thing I noticed is that when calculating congestion_length, it also calculates the total length of uncongested_sequence, even though the length is defined as 0. Could we avoid that calculation because for most of time period, most contiguous stations are uncongested. I suppose that would consume notable time to calculate the sum function for them (please correct me if I'm wrong). I will continue checking the codes on Tuesday, especially on the bottleneck logic compared with the paper and making necessary comments.

This feedback was addressed live on a call with Hang and Ian. We agreed that the current model logic addresses not summing zero lengths.

thehanggit:
@ian-r-rose @britt-allen I compared the inverse way to calculate the bottleneck by switching the logic from ascending postmile to descending postmile:
image
image
The results seem to stay same since the basic logic (v_downstream - v_upstream) is not changed. Consequently, no matter from which direction I am sorting (either from start point of a route or end point of a route), the bottleneck position will be fixed. Let me know what's your thought and we can have a discussion based on it.

ian-r-rose:
@thehanggit I think if you swap `lead` and `lag`, and change `asc` to `desc` in the window functions, you've actually made no changes to the data.

I think you'll only want to make one of those changes to the window function, and make sure you get the sign of the speed delta right so that you are looking at the end of congestion rather than the start. I *do* expect there to be differences in the two different approaches to bottlenecks, largely having to do with the stability of the starting/exiting station

thehanggit:
@ian-r-rose Thank you for the instructions! Let me try to clarify the different issues here. By bottleneck definition, if I only swap `lead` and `lag` in calcs subquery and change the speed diff criteria below:
image

Then the bottleneck would be defined as upstream station, which has higher speed (e.g. 60 mph). It may not make sense to define bottleneck start upstream. But generated results are indeed different from what we have done before.

_As you mention by looking at the end of congestion rather than the start_, should we just change subqueries of  `congestion_events` and `congestion_length`? For `congestion_events`, we could sequence events in a descending order of postmile to define downstream station as start of a new congestion event. Then do similar thing for `congestion_length` to calculate the length. 

I believe these two changes affect the stability of starting/exiting station.

thehanggit:
@britt-allen @ian-r-rose Since I'm already familiar with the bottleneck model, I can try to finish the qualify is_bottleneck and get it ready for your review.

britt-allen:
> @britt-allen @ian-r-rose Since I'm already familiar with the bottleneck model, I can try to finish the qualify is_bottleneck and get it ready for your review.

Thank you @thehanggit

transform/models/intermediate/performance/int_performance__bottlenecks.sql
```diff
@@ -24,6 +24,25 @@ station_five_minute as (
         and station_type in ('ML', 'HV')
 ),
 
+station_config as (
+    select
+        station_id,
+        max(latitude) as lat,
+        min(longitude) as long
```
britt-allen, line 31: Lat and Long already exist further upstream in `int_clearinghouse__detector_agg_five_minutes`. It may make since to include those in the downstream models of `int_clearinghouse__detector_agg_five_minutes` that power the bottlenecks model rather than doing a join here if these are columns that will be commonly used in other analyses. I'll defer to @kengodleskidot and maybe @JamesSLogan or @tnrahim can weigh in on performance implications of either path.

thehanggit: @britt-allen Thank you Britt for the review! Yes, I think long and lat would be commonly used in other performance metrics for visualization. I will talk with @kengodleskidot tomorrow to settle it.

thehanggit:
@britt-allen Visualized the bottleneck congestion on 15:00 8/14/2024. Right now the congestion is on segment level not for station level, meaning one segment may contains multiple stations. I will breakdown segments into station-based pieces and test our congestion logic through visualization

image

thehanggit (approved):

Issue #311, opened 2024-07-18
Create mart for imputation report


ian-r-rose:
Now that we have our imputation methods mostly implemented, we should create a summary table based on the one in #297 . For each station and date, we could have:

* % observed
* % of each imputation method
* % unobserved and unimputed (i.e., values that are still null after all our attempts to impute)

This table could be further aggregated in a BI tool to freeway, district, county, etc.

@mmmiah would you be interested in taking this one on?

mmmiah:
Yes, I can work on this

mmmiah:
closed as completed

PR #324, opened 2024-07-23
Imputation mart model

mmmiah:
This model provides the summary of imputation by different methods. The model calculated the percentage of speed, volume and occupancy imputed by different methods as well as reported percentage of observed and unobserved/unimputed. This PR is build according the issue #311


transform/models/marts/imputation/imputation__station_summary.sql
```diff
@@ -0,0 +1,71 @@
...
+
+imputation_count as (
+    select
+        station_id,
+        district,
+        sample_date,
+        direction,
+        lane,
+        freeway,
+        count(station_id) as sample_ct,
```
ian-r-rose, line 22: I think it's more legible to use `count(*)` here unless you have a reason for specifically counting station ids.

transform/models/marts/imputation/imputation__station_summary.sql
```diff
@@ -0,0 +1,71 @@
+{{ config(
+    materialized="incremental",
+    unique_key=['station_id', 'sample_date'],
+    snowflake_warehouse=get_snowflake_warehouse(size="XL")
```
ian-r-rose, line 4: For incremental models, let's use the `get_snowflake_refresh_warehouse` macro. That one has specific behavior where it goes to a larger warehouse during full refreshes.

transform/models/marts/imputation/imputation__station_summary.sql
```diff
@@ -0,0 +1,71 @@
...
+        count(case when volume_imputation_method = 'local' then station_id end) as vol_local_imputation_sample,
+        count(case when volume_imputation_method = 'regional' then station_id end) as vol_regional_imputation_sample,
+        count(case when volume_imputation_method = 'global' then station_id end) as vol_global_imputation_sample,
+        count(case when volume_imputation_method = 'observed' then station_id end) as vol_observed_sample,
+        count(case when volume_imputation_method is NULL then station_id end) as vol_unobserved_unimputed,
+        count(case when speed_imputation_method = 'local' then station_id end) as speed_local_imputation_sample,
+        count(case when speed_imputation_method = 'regional' then station_id end) as speed_regional_imputation_sample,
+        count(case when speed_imputation_method = 'global' then station_id end) as speed_global_imputation_sample,
+        count(case when speed_imputation_method = 'observed' then station_id end) as speed_observed_sample,
+        count(case when speed_imputation_method is NULL then station_id end) as speed_unobserved_unimputed
```
ian-r-rose, line 37: I think most of these could use `count_if` to be a bit more legible, and avoid the need for an internal case statement, e.g.

```sql
count_if(occupancy_imputation_method = 'local') as occ_local_imputation_sample
```

transform/models/marts/imputation/imputation__station_summary.sql
```diff
@@ -0,0 +1,71 @@
...
+    from obs_imputed_five_minutes_agg
+    group by station_id, district, lane, direction, freeway, sample_date
```
ian-r-rose, line 39: Can you just group by `station_id`, `lane`, and `sample_date`, and use `any_value()` for the other metadata fields? I think it's more legible to keep the group by keys to just the things we expect to be unique for the table, and the incidental denormalized metadata fields are just along for the ride. For example, here:

https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/c04a61a74992215ab0b9d4d43454904c68ed821a/transform/models/intermediate/performance/int_performance__detector_metrics_agg_hourly.sql#L17-L56

transform/models/marts/imputation/imputation__station_summary.sql
```diff
@@ -0,0 +1,71 @@
...
+imputation_pct as (
+    select
+        station_id,
+        district,
+        sample_date,
+        direction,
+        lane,
+        freeway,
+        sample_ct,
+        (coalesce(occ_local_imputation_sample, 0) / nullif(sample_ct, 0)) * 100 as pct_of_occupancy_local_regression,
```
ian-r-rose, line 51: Here and below, rather than using `nullif(sample_ct, 0)`, you can use the dedicated function for it `nullifzero`

transform/models/marts/imputation/imputation__station_summary.sql
ian-r-rose, line 1: Since this is a new subdirectory in the `marts` folder, can you add a schema specification here? https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/c04a61a74992215ab0b9d4d43454904c68ed821a/transform/dbt_project.yml#L85-L86

This will make sure it's built in the right schema, as well as the right folder in the S3 bucket

ian-r-rose: Note that this table is automatically being built in the dev marts S3 bucket [here](https://caltrans-pems-dev-us-west-2-marts.s3.us-west-2.amazonaws.com/dbt_cloud_pr_15_324/station_summary.parquet), which you can use to test PowerBI or AGOL dashboards.

ian-r-rose (commented):
Thanks @mmmiah, this is looking great! I have a few minor suggestions around legibility, but it's looking very close to me.

transform/models/marts/imputation/imputation__station_summary.sql
mmmiah: Alright changed to nullifzero function

transform/models/marts/imputation/imputation__station_summary.sql
mmmiah: updated the aggregated factors to as 'group by station_id, lane, sample_date'

transform/models/marts/imputation/imputation__station_summary.sql
mmmiah: replaced based on your suggestions

transform/models/marts/imputation/imputation__station_summary.sql
mmmiah: Sure! Thanks for suggesting it!

transform/models/marts/imputation/imputation__station_summary.sql
mmmiah: updated

transform/models/marts/imputation/imputation__station_summary.sql
```diff
@@ -0,0 +1,64 @@
+{{ config(
+    materialized="incremental",
+    unique_key=['station_id', 'sample_date'],
```
ian-r-rose, line 3: I think `lane` is also a unique key for this model, which is at the detector leve.

transform/models/marts/imputation/imputation__station_summary.sql
ian-r-rose, line 1: If I'm reading this correctly, this model is at the detector level, so I think this might be better named `imputation__detector_summary`.

ian-r-rose (commented):
Looks good to me, just a very minor comment about the model grain

transform/models/marts/imputation/imputation__station_summary.sql
mmmiah: sure, renamed as 'imputation__detector_summary'

transform/models/marts/imputation/imputation__station_summary.sql
mmmiah: agreed, included the lane

mmmiah:
@ian-r-rose , could you please see why I am getting dbt error for this PR. I tried to delete all PR related to performance, still facing dbt CI error

ian-r-rose:
It looks like it was running into troubles with dbt_cloud_pr_15_324_imputation. I dropped that schema and tried re-running, let's see how it does

mmmiah:
@ian-r-rose , it passed the dbt CI

ian-r-rose (approved):
In it goes!

PR #329, opened 2024-07-26
Detector status report

### LINKED ISSUE #269: Create detector status report
Based on `int_diagnostics__detector_status`, we should create a human-readable status report which aggregates it up to either the district, or freeway-district combination (or a version of both). This would include things like:

* Number of stations with each status
* Fraction of stations with "good" status
* ...
* ...

cc @kengodleskidot who might have better ideas for nice summary statistics

kengodleskidot:
Updates the diagnostic models to use the latest configuration data and creates two new reports in the marts folder for BI tool and reporting usage.  Fixes #269.


kengodleskidot:
Created a new branch to perform this work and is not included in #332

Issue #273, opened 2024-07-05
Use status column in config tables to handle decomissioning of stations/controllers/detectors


ian-r-rose:
The config log tables have a "status" column, which seemingly indicates when a station has been decomissioned. Right now, we are not using that column in our SCD-2 tables, so our metadata tables have some devices which are not currently active.

We should use the "status" column to determine when a station is no longer active and ensure that those stations do not show up in the final tables (or are easy to filter out)

kengodleskidot:
Thanks for bringing this up @ian-r-rose and I will check the status column in our new tables against what I see in the current PeMS to confirm your observation initially mentioned.

ian-r-rose:
@britt-allen pending the results of @kengodleskidot's investigation, would you be able to take a look at this?

britt-allen:
I will be able @ian-r-rose

britt-allen:
@kengodleskidot can you share the results of your investigation, please?

PR #332, opened 2024-07-27
Detector Status Updates and New Diagnostic Mart tables

### LINKED ISSUE #269: Create detector status report

kengodleskidot:
Updates the diagnostic models to use the latest configuration data and creates two new reports in the marts folder for BI tool and reporting usage. Fixes #269 .


transform/models/intermediate/clearinghouse/int_clearinghouse__active_detectors.sql
ian-r-rose, line 1: This is very nice and succinct

transform/models/intermediate/diagnostics/int_diagnostics__det_diag_set_assignment.sql
ian-r-rose, line 1: I expected this model to start selecting from your new `int_clearinghouse__active_detectors.sql`. Or is the diagnostic set only meaningful at the station level?

transform/models/intermediate/clearinghouse/int_clearinghouse__active_detectors.sql
ian-r-rose: Now that this data no longer depends on anything from the "clearinghouse" feeds, can we move it the `vds` directory and call it `int_vds__active_detectors.sql`? I'm trying to de-emphasize clearinghouse for models that aren't tightly coupled to those feeds

transform/models/marts/diagnostics/diagnostics__detector_daily_by_station.sql
ian-r-rose, line 1: This is looking great! Since this is a new subdirectory of "marts", can you specify a schema in the `dbt_project.yml`?

Note that this file is now available in the dev marts S3 bucket [here](https://caltrans-pems-dev-us-west-2-marts.s3.us-west-2.amazonaws.com/dbt_cloud_pr_15_332/detector_daily_by_station.parquet). You can use the dev CI builds of the files to test any PowerBI or AGOL visualizations you might want to experiment.

transform/models/marts/diagnostics/diagnostics__detector_daily_by_station.sql
```diff
@@ -0,0 +1,66 @@
...
+detector_status_with_count as (
+    select
+        district,
+        station_id,
+        lane,
+        station_type,
+        sample_date,
+        sample_ct,
+        count_if(status = 'Good') as good_detector,
+        count_if(status != 'Good') as bad_detector
```
ian-r-rose, line 18: Is there a reason to have this as two columns? I might instead have one column for good detectors, and another for total detector count. Bad detectors would then be total - good.

transform/models/marts/diagnostics/diagnostics__detector_daily_by_station.sql
```diff
@@ -0,0 +1,66 @@
...
+    group by district, station_id, lane, station_type, sample_date, sample_ct
+),
+
+detector_status_by_station as (
+    select
+        district,
+        station_id,
+        station_type,
+        sample_date,
+        count(lane) as detector_count,
```
ian-r-rose, line 29: I found this a bit confusing, wouldn't `count(*)` do the same thing? Or is it because `lane` is null for inactive stations (cf #273)? Would we be able to just do `count(*)` if #273 is fixed?

transform/models/marts/diagnostics/diagnostics__detector_daily_detail.sql
ian-r-rose, line 1: Just musing: should we make sure that these metadata are on the `int_diagnostics__detector_status` model? They are now mostly available on its base model, so it would not be a huge lift, and would avoid needing to do the join here.

ian-r-rose (commented):
This is really great work @kengodleskidot, I have a few very minor suggestions, but I think this is looking just about ready to go in.

transform/models/intermediate/diagnostics/int_diagnostics__det_diag_set_assignment.sql
kengodleskidot: The diagnostic set is only meaningful at the station level since the threshold set used is based on district and station type. The detector data would contain repetitive station values that are not necessarily needed. 
I created and was going to join `int_clearinghouse__active_detectors` with the `int_diagnostic__detector_status` model but saw that Taheer was working on something similar so I held off.

transform/models/marts/diagnostics/diagnostics__detector_daily_by_station.sql
kengodleskidot: I agree that count(*) would do the same thing and will update the code accordingly. I am seeing a small discrepancy in station counts when using the status column in the VDS config files but will respond directly in #273 with my results.

transform/models/marts/diagnostics/diagnostics__detector_daily_detail.sql
kengodleskidot: If the metadata can be added to the `int_diagnostics__detector_status` without affecting performance, we would be able to create the marts tables more easily.

transform/models/intermediate/diagnostics/int_diagnostics__det_diag_set_assignment.sql
ian-r-rose: Okay, makes sense!

transform/models/marts/diagnostics/diagnostics__detector_daily_detail.sql
ian-r-rose: I think it could be added there without affecting performance, since the metadata are already on `int_clearinghouse__detector_agg_five_minutes`

transform/models/marts/diagnostics/diagnostics__detector_daily_by_station.sql
kengodleskidot: Most of the reports display a count for good and bad detectors which is why I show them as separate columns. Also, most users tend to focus on one or another (Good or Bad) and then drill into the Bad detectors to find the reason. In the columns you mention, would the Bad detector count be done as a measure in the visualization tool or should I have a third column for all 3 (total, good and bad)?

ian-r-rose: That makes sense to me, I was just looking for opportunities to simplify the table. But you have a better idea of the intended visualizations than I do, so whatever you think would be most helpful should be good!

transform/models/marts/diagnostics/diagnostics__detector_daily_detail.sql
kengodleskidot: I would like to update this in a follow up PR since there is discussion about including the lat/long as a geospatial point object. I will create an issue for it and assign it to myself.

tnrahim:
Hi @kengodleskidot, 

Things look good to me!

ian-r-rose (approved):
Looks great, thanks @kengodleskidot!

PR #335, opened 2024-07-29
GIS file for testing_ArcGIS_Enterprise_feasibility

mmmiah:
This is the PR we generated to create a test geospatial file to understand the feasibility of ArcGIS enterprise to make live connection with snowflake mart database.


transform/models/marts/performance/test_detectors_health_GIS_file.sql
```diff
@@ -0,0 +1,44 @@
...
+        GL.DIRECTION,
+        GL.ABSOLUTE_POSTMILE
+    FROM DETECTOR_HEALTH AS DH
+    INNER JOIN GEO_LOC AS GL ON DH.STATION_ID = GL.STATION_ID
+),
+
+DETECTOR_HEALTH_WITH_GEOGRAPHY AS (
+    SELECT
+        *,
+        TO_GEOGRAPHY('POINT(' || LONGITUDE || ' ' || LATITUDE || ')') AS GEOGRAPHY
```
ian-r-rose, line 39: [ST_POINT](https://docs.snowflake.com/en/sql-reference/functions/st_makepoint) should be a better way to construct geography that doesn't rely on string concatenation

ian-r-rose: How are you trying to read the table? Using the ODBC connector, or from S3?

transform/models/marts/performance/test_detectors_health_GIS_file.sql
```diff
@@ -0,0 +1,44 @@
...
+
+WITH DETECTOR_HEALTH AS (
+    SELECT DISTINCT
+        D.STATION_ID,
+        D.DISTRICT,
+        D.TYPE AS DETECTOR_TYPE,
+        D.LANE,
+        D.SAMPLE_DATE,
+        D.STATUS
+    FROM TRANSFORM_PRD.DIAGNOSTICS.INT_DIAGNOSTICS__REAL_DETECTOR_STATUS AS D
```
ian-r-rose, line 11: This should be `{{ ref("int_diagnostics__real_detector_status") }}`

transform/models/marts/performance/test_detectors_health_GIS_file.sql
```diff
@@ -0,0 +1,44 @@
...
+),
+
+GEO_LOC AS (
+    SELECT DISTINCT
+        V.STATION_ID,
+        V.LATITUDE,
+        V.LONGITUDE,
+        V.DIRECTION,
+        V.ABSOLUTE_POSTMILE
+    FROM TRANSFORM_PRD.VDS.INT_VDS__DETECTOR_CONFIG AS V
```
ian-r-rose, line 22: Also here

transform/models/marts/performance/test_detectors_health_GIS_file.sql
```diff
@@ -0,0 +1,44 @@
...
+WITH DETECTOR_HEALTH AS (
+    SELECT DISTINCT
+        D.STATION_ID,
+        D.DISTRICT,
+        D.TYPE AS DETECTOR_TYPE,
+        D.LANE,
+        D.SAMPLE_DATE,
+        D.STATUS
+    FROM TRANSFORM_PRD.DIAGNOSTICS.INT_DIAGNOSTICS__REAL_DETECTOR_STATUS AS D
+    WHERE D.SAMPLE_DATE = '2024-07-18'
```
ian-r-rose, line 12: Can we future-proof this date? What is the intention of this table?

transform/models/marts/performance/test_detectors_health_GIS_file.sql
ian-r-rose, line 1: I don't really like naming this model a "file", since it's a table in Snowflake, rather than a file at this point

ian-r-rose (commented):
I'm not really sure about the purpose of this model. Isn't this largely the same data that is in the model @kengodleskidot is working on in #332 ? If so, it would make more sense to me to combine your efforts.

Alternatively, we could create a more targeted table that just contains detector or station metadata, and publish it to S3 as GeoJSON (I have some experiments along these lines).

But overall, I want to make sure you can test these things in dev to ensure they satisfy your requirements. Right now, I don't think we need to merge a test model to prod to do any testing, we just need to make sure you can connect to the dev marts using ArcGIS.

mmmiah:
@ian-r-rose we do not need to be this table perfect for everything, this is sample one day data we are testing from mart. We can delete this table later once, we are done with the feasibility test. None of the file in mart folder is readable now due to lack of geometry property. We need to complete this feasibility test by this week before having meeting with contractor, S3 bucket connection is more complex and need python script to connect. We are very close to test with ArcGIS to snowflake mart. We already registered the snowflake mart with ArcGIS enterprise to test it. I do not know we have time to follow the S3 bucket route now in GIS, however, we may do that in future

transform/models/marts/performance/test_detectors_health_GIS_file.sql
mmmiah: using ODBC connection in ArcGIS Pro

transform/models/marts/performance/test_detectors_health_GIS_file.sql
mmmiah: just testing the real time connection with ArcGIS enterprise and real time update of the data based on user query id.

ian-r-rose:
> @ian-r-rose we do not need to be this table perfect for everything, this is a one day sample data we are testing from mart. We can delete this table later once, we are done with the feasibility test. None of the file in mart folder is readable now due to lack of geometry property. We need to complete this feasibility test by this week before having meeting with contractor, S3 bucket connection is more complex and need python script to connect. We are very close to test with ArcGIS to snowflake mart. We already registered the snowflake mart with ArcGIS enterprise to test it. I do not know we have time to follow the S3 bucket route now in GIS, however, we may do that in future

One of the key features of the workflow we are trying to develop here is that we want to provide safe ways to test things in dev while not changing the production pipelines. While it's true that it's not a huge deal to add and then later delete a test table, I think it is much more valuable to be able to use a dev pipeline effectively. Can you connect to the `ANALYTICS_DEV` database using the `REPORTER_DEV` role and `REPORTING_XS_DEV` warehouse? The table built by this PR *should already be accessible* using those parameters, and without having to merge anything to production. It doesn't require anything to be written to S3, either (though I do have some work-in-progress there that I think we should be able to complete in the next week)

mmmiah:
@ian-r-rose , Alright,I think that we need to reach again to ArcGIS Eterprise administrator to see if they can register the dev connection for us

ian-r-rose:
> @ian-r-rose , Alright,I think that we need to reach again to ArcGIS Eterprise administrator to see if they can register the dev connection for us

I don't think you need to do this, you just need to create a new connection to Snowflake, in the way I saw you do last week.

mmmiah:
@ian-r-rose Right now only the production is registered in ArcGIS Enterprise to store the data through connection. this is different than what you saw in last week ArcGIS Pro desktop connection. I will do my homework and see if it works. Anyway, Thanks!

ian-r-rose:
Would you be willing to screenshare? I think it's important to be able to test in dev instead of prod (that's why we have a dev environment!), so if there are tweaks we need to make, it would be good to understand them

mmmiah:
@ian-r-rose @kengodleskidot  , yes, sure. Here is the step actually we are following to make data live from snowflake to ArcGIS Enterprise-
 Step 1. Connecting snowflake specific database with ArcGIS Pro and creating sde file which is very specific for each environment as shown below-

image

Step 2. Send the sde file to Enterprise administration to register, otherwise your data will not be live. You can see the screen shoot for the requirement of registration-
image

You can see that this sde file is very specific as shown below
image
If you use the registered sde file for publishing that data will be live otherwise not.

No worries, I have connected the annalytical dev folder and will work administrative people to register dev as well. However, we need to work on geospatial data for Analytical PRD

ian-r-rose:
So, what we can do is create two different service accounts, one for dev and one for prod, and register both. I think that would be the best way to support both the development of new datasets as well as the use of production datasets in stable dashboards.

How difficult would that be to set up from your perspective?

mmmiah:
@ian-r-rose , @kengodleskidot ,
Here you can see the test, the enterprise is saying your sde is not registered and can not publish error
image
 yes, I am going to request them to do that.

ian-r-rose:
Let me set you up with some separate credentials for a dev service account to keep things separate before you do

mmmiah:
will there be any change in sde file input as shown above for Dev, then I will wait for you before sending request to Enterprise admins? If not, I will send them request. Please let me know. Thank you for heads up!

ian-r-rose:
It will be the same, but with the username `AGOL_SVC_USER_DEV`, and a separate password

mmmiah:
Ok, then I will wait for you to get the new login info, will test ArcGIS pro connection and then generate sde file for admins

ian-r-rose:
Just sent you some new credentials

mmmiah:
Connection failed, is there any other change
image

ian-r-rose:
I tried cycling the password and sent you a new link

mmmiah:
@ian-r-rose , it works, Thanks

PR #342, opened 2024-07-31
Include Normalized Flow in Detector 5-min agg Model

kengodleskidot:
Create a `volume_normalized` field in the `int_clearinghouse__detector_agg_five_minutes.sql` model that returns a volume value based on the number of 30-second samples received and update the associated yml. For example, if a detector has a volume of 40 vehicles in 5 minutes but only 5 out of 10 samples are received (5 samples are not reported by the device), the normalized value is 10 samples / 5 samples recieved  * 40 vehicles = 80 vehicles

### DIFF
--- a/transform/models/intermediate/clearinghouse/_clearinghouse.yml
+++ b/transform/models/intermediate/clearinghouse/_clearinghouse.yml
@@ -107,6 +107,12 @@ models:
           Lane associated with a route and station from raw data.
       - name: volume_sum
         description: Number of vehicle that passed over the detector during the sample period for the lane.
+      - name: volume_normalized
+        description: |
+          For the 5-minute aggregation there should be 10 30-second samples collected. If 10 or more
+          samples are collected the aggregated (sum) volume rounded to the nearest integer value is used
+          but if less than 10 samples are recieved in the 5-minute timeframe the flow is normalized using
+          the equation: 10 / sample count * sum(volume)
       - name: zero_vol_ct
         description: Counts the number of raw data samples where a lane's volume (flow) value equals 0.
       - name: occupancy_avg
--- a/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
+++ b/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
@@ -83,6 +83,11 @@ agg as (
             sample_ct_{{ lane }} as sample_ct,
             {{ lane }} as lane,
             volume_{{ lane }} as volume_sum,
+            round(iff(
+                sample_ct_{{ lane }} >= 10, volume_{{ lane }},
+                10 / nullifzero(sample_ct_{{ lane }}) * volume_{{ lane }}
+            ))
+                as volume_normalized,
             zero_vol_ct_{{ lane }} as zero_vol_ct,
             occupancy_{{ lane }} as occupancy_avg,
             zero_occ_ct_{{ lane }} as zero_occ_ct,


transform/models/intermediate/clearinghouse/_clearinghouse.yml
thehanggit, line 1: The normalized aggregation looks good to me. It works if 30-second samples are nearly evenly distributed (Since 5-minute traffic volumes do not fluctuate much). Here are my additional comments:

1. In which circumstances would there be 10 more samples in the 5-minute aggregation?
2. Should we simply use the imputed data to avoid missing samples?
3. I can apply more advanced models (ARIMA, SARIMA, GARCH) for time series analysis

thehanggit (commented):
@kengodleskidot 
The normalized aggregation logic and code look good to me. It works if 30-second samples are nearly evenly distributed (Since 5-minute traffic volumes do not fluctuate much). Here are my additional comments:

1. In which circumstances would there be 10 more samples in the 5-minute aggregation?
2. Should we simply use the imputed data to avoid missing samples?
3. I can apply more advanced models (ARIMA, SARIMA, GARCH) for time series data if necessary

mmmiah (approved):
The logic seems good to me! Just curious, did you get chance to investigate the extreme case, I mean what percentage of sample have more than 10, 30-sec samples in the historical period of time.

thehanggit (approved):

ian-r-rose (approved):
This looks good to me @kengodleskidot, thanks!

Is it worth using `volume_normalized` in some of the downstream models here? Or do you want to do that in a follow-up?

I'm also curious about @thehanggit's question: do you have a sense of how common it is for detectors to have fewer than 10 samples in 5 minutes? Is it, e.g., at the 10% level? 1% level?

kengodleskidot:
> @kengodleskidot The normalized aggregation logic and code look good to me. It works if 30-second samples are nearly evenly distributed (Since 5-minute traffic volumes do not fluctuate much). Here are my additional comments:
> 
> 1. In which circumstances would there be 10 more samples in the 5-minute aggregation?
> 2. Should we simply use the imputed data to avoid missing samples?
> 3. I can apply more advanced models (ARIMA, SARIMA, GARCH) for time series data if necessary

@thehanggit below are my replies to the questions above:
1. We have some ramp stations that are reporting up to 30 samples in a 5-minute timespan. Based on my previous analysis these stations were mainly located in districts 3 and 7.
2. I am not sure that the use of imputed data is any more accurate than using normalized flow data and the normalization is only related to flow. This would also require more compute resources and complicated code. I do see the need for putting an upper limit on the normalized flow value but that would be part of the final 5-minute performance model.
3. Thank you for the offer and we can look at exploring this in the future 

FYI below is a quick analysis I performed regarding sample count. I took the 5-minute agg model from the clearinghouse for 7/1/24 and 7/31/24 and below are the counts and percentages of the sample distribution based on the samples recieved:
 
image

kengodleskidot:
> This looks good to me @kengodleskidot, thanks!
> 
> Is it worth using `volume_normalized` in some of the downstream models here? Or do you want to do that in a follow-up?
> 
> I'm also curious about @thehanggit's question: do you have a sense of how common it is for detectors to have fewer than 10 samples in 5 minutes? Is it, e.g., at the 10% level? 1% level?

Below is a quick analysis I performed regarding sample count. I took the 5-minute agg model from the clearinghouse for 7/1/24 and 7/31/24 and below are the counts and percentages of the sample distribution based on the samples recieved:
image

PR #345, opened 2024-08-01
Bump dbt-core version

ian-r-rose:
This bumps the version of dbt specified in our pyproject.toml to match what is in our cloud environment


britt-allen (approved):
Looks great, thank you for the quick fix and for feeding two birds with one seed.

PR #346, opened 2024-08-01
Update big refresh warehouse to XL

### LINKED ISSUE #333: Change refresh warehouse macro defaults
Now that most of our models have better performance characteristics (cf #211 ), we should be able to remove all usage of 4XL warehouses from this project and save a little money. Most usage of 4XL is done incidentally right now due to the default size in the `get_snowflake_refresh_warehouse` macro:

https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/b3ce9c19e0d8ebac00eef2a3be1171b1c537d9f9/transform/macros/get_snowflake_refresh_warehouse.sql#L1

Let's change the defaults to be XS for the small warehouse, and XL for the big warehouse.

kengodleskidot:
The default for big warehouse updated to XL from 4XL. Small warehouse remains as XS. Fixes #333


ian-r-rose (approved):
Thanks @kengodleskidot!

britt-allen (approved):
Thank you Ken!

Issue #347, opened 2024-08-02
Remove `int_diagnostics__real_detector_status` model


ian-r-rose:
Now that we have more reliable metadata in `int_vds__detector_config`, and our five-minute-agg model only includes real lanes (
#214) the `int_diagnostics__real_detector_status` model is no longer needed. It was always intended to be temporary, and can be removed all-together

PR #348, opened 2024-08-02
Clean up regression coefficient models

ian-r-rose:
I noticed that the model for regression coefficients had a performance regression (in the other sense). In looking into it, I found a bad join with the `nearby_stations` model, which meant that we were getting too many "nearby" stations in our regression because it wasn't taking the date into account.

I also did a bit of clean up in those files, including fixing #347


JamesSLogan (approved):
Looks good - thanks Ian!

PR #349, opened 2024-08-05
updated_logic to tag the imputation value, method and dates

### LINKED ISSUE #326: Bugs in imputation logic
1. There is a bug in the imputation_method column for the `int_imputation__detector_imputed_agg_five_minutes` model. Right now it's reporting observed for every row when it's clear that some have been imputed.

```sql
select VOLUME_IMPUTATION_METHOD, count(*) as a_count
from transform_prd.imputation.int_imputation__detector_imputed_agg_five_minutes
where "STATION_ID" = '826789'
and "SAMPLE_DATE" = '2024-07-17'
group by VOLUME_IMPUTATION_METHOD
```

2. Speed is 0 when volume and occupancy are non-zero, we should be imputing for speed in this case but we are not

```sql
select *
from transform_prd.imputation.int_imputation__detector_imputed_agg_five_minutes
where "STATION_ID" = '826789'
and "SAMPLE_DATE" = '2024-07-17'
and "OCCUPANCY_AVG" > 0 and "VOLUME_SUM" > 0 and "SPEED_FIVE_MINS" = 0
```

Here's a [snowflake worksheet](https://app.snowflake.com/vsb79059/dse_caltrans_pems/w4BKFk3md5jB#query) to reproduce both 1 and 2 as well, the code in the worksheet is the same as the snippets above.

mmmiah:
This PR updated the imputation logic for the following case based on Fixes #326

1. Detector_is_good=false, impute volume, occupancy and speed
2. Detector_is_good=True, volume_sum is null, impute volume_sum
3. Detector_is_good=True, occupancy_avg is null, impute occupancy_avg
4. Detector_is_good=True, speed_five_mins is null, impute_five_mins_speed

### DIFF
--- a/transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
+++ b/transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
@@ -37,19 +37,19 @@ hybrid_five_mins_agg as (
         station_valid_to,
         -- select the imputed value
         case
-            when detector_is_good = false
+            when detector_is_good = false or (detector_is_good = true and volume_sum is null)
                 then
                     coalesce(volume_local_regression, volume_regional_regression, volume_global_regression)
             else volume_sum
         end as volume_sum,
         case
-            when detector_is_good = false
+            when detector_is_good = false or (detector_is_good = true and speed_five_mins is null)
                 then
                     coalesce(speed_local_regression, speed_regional_regression, speed_global_regression)
             else speed_five_mins
         end as speed_five_mins,
         case
-            when detector_is_good = false
+            when detector_is_good = false or (detector_is_good = true and occupancy_avg is null)
                 then
                     coalesce(
                         occupancy_local_regression,
@@ -60,7 +60,11 @@ hybrid_five_mins_agg as (
         end as occupancy_avg,
         -- assign the imputation date
         case
-            when detector_is_good = false
+            when
+                detector_is_good = false
+                or (detector_is_good = true and volume_sum is null)
+                or (detector_is_good = true and occupancy_avg is null)
+                or (detector_is_good = true and speed_five_mins is null)
                 then
                     coalesce(local_regression_date, regional_regression_date, global_regression_date)
             else sample_date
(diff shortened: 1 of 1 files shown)


transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
```diff
@@ -37,19 +37,19 @@ hybrid_five_mins_agg as (
         station_valid_to,
         -- select the imputed value
         case
-            when detector_is_good = false
+            when detector_is_good = false or (detector_is_good = true and volume_sum is null)
```
JamesSLogan, line 40: This could be simplified to `when detector_is_good = false or volume_sum is null`.

Same goes for the `speed_five_mins` and `occupancy_local_regression` below.

transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
```diff
@@ -60,7 +60,11 @@ hybrid_five_mins_agg as (
         end as occupancy_avg,
         -- assign the imputation date
         case
-            when detector_is_good = false
+            when
+                detector_is_good = false
+                or (detector_is_good = true and volume_sum is null)
+                or (detector_is_good = true and occupancy_avg is null)
+                or (detector_is_good = true and speed_five_mins is null)
```
JamesSLogan, line 67: This can be simplified to match the above logic, i.e.
```sql
when detector_is_good = false 
  or volume_sum is null 
  or occupancy_avg is null 
  or speed_five_mins is null
```

transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
```diff
@@ -71,33 +75,57 @@ hybrid_five_mins_agg as (
             when
                 detector_is_good = false and volume_local_regression is not null
                 then 'local'
+            when
+                detector_is_good = true and volume_sum is null and volume_local_regression is not null
+                then 'local'
```
JamesSLogan, line 80: I think it would be clearer to update the logic within the existing `when` clauses rather than to add new ones.

```sql
...
when
    (detector_is_good = false or volume_sum is null) and volume_local_regression is not null
    then 'local'
when
     (detector_is_good = false or volume_sum is null) and volume_regional_regression is not null
    then 'regional'
...
```

Same goes for following lines.

JamesSLogan (changes requested):
Thanks Mintu! The core logic looks good, I added a few comments around simplification and readability.

Could you add the text "Fixes #326" in the description to link this PR to the issue?

transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
mmmiah: great suggestions! Agreed!

transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
mmmiah: updated!!!

transform/models/intermediate/imputation/int_imputation__detector_imputed_agg_five_minutes.sql
mmmiah: good catch James, just updated based on your suggestions

JamesSLogan (approved):

PR #351, opened 2024-08-05
Geo data loading

### LINKED ISSUE #233: Caltrans GIS Data - Load datasets
This post is informational about the GIS related files available and managed by Caltrans which can be found at the following link: https://gisdata-caltrans.opendata.arcgis.com/search?tags=Highway

As we look to incorporate GIS data into PeMS, there are two datasets that are of particular interest under the Highway Datasets section:
- All Roads (https://gisdata-caltrans.opendata.arcgis.com/maps/2d56e65de89c418780056651640291e8): The Caltrans All Roads Linear Referencing Services (LRS) dataset provides the base geometry for federally required Highway Performance Monitoring System (HPMS) business data, functionally classified roads for the California Roads System (CRS) (a requirement for federal funding of local agency projects), and the State Highway Network (SHN), which supports a wide range of internal Caltrans business needs.
- SHN Postmiles Tenth (https://gisdata-caltrans.opendata.arcgis.com/maps/c22341fec9c74c6b9488ee4da23dd967): Postmiles are valid postmile points at 0.1 (1/10th) mile intervals, based on the Caltrans Linear Referencing System. Exceptions that are included are the beginning and ends of routes (and other situations) that don't necessarily begin or end on the even 0.1 mile interval. This geometry is based upon the extraction from TSN on 10 October 2022. 

Other GIS data sets include district, county, legislative boundaries as well as a bottleneck data set. As other GIS data is encountered this issue should be updated accordingly as needed.

britt-allen:
closes #233


britt-allen:
This is a work in progress and does not yet require feedback.

transform/models/_sources.yml
```diff
@@ -408,3 +408,136 @@ sources:
           - name: slot
           - name: volume_flag
           - name: logical_position
+  - name: geo_reference
+    database: RAW_DEV
```
ian-r-rose, line 412: We'll be loading this into `RAW_PRD`, so let's document it that way

ian-r-rose (commented):
Looks good to me @britt-allen, just one minor comment

PR #355, opened 2024-08-08
Unload geospatial marts as geojson

ian-r-rose:
This PR creates a new stored procedure `unload_as_geojson()` which takes a geospatial table and writes it to our marts S3 bucket as GeoJSON. Any mart model which is in the `marts/geo` directory has this stored procedure called for it.

The intention here is to use it to automatically unload geospatial mart tables to a format that can be consumed by ArcGIS. Note that GeoJSON is *not* a very compact format, and I recommend only including tables with fewer than about ~1M rows in the `marts/geo` directory. Otherwise, things will likely start to fail.


ian-r-rose:
@britt-allen I wrote up a quick github action to create the stored procedures here, and included the geospatial data loads while I was at it.

britt-allen (approved):
Thank you Ian, looks good to me!

PR #356, opened 2024-08-08
Move profile config to avoid deprecation warning

ian-r-rose:
Docs build started [failing](https://github.com/cagov/caldata-mdsa-caltrans-pems/actions/runs/10307510120/job/28532925162) because dbt deprecated configuration in the `profiles.yml`. This moves it to the new location in `dbt_project.yml`


britt-allen (approved):
Thank you for catching and fixing this so quickly!

PR #357, opened 2024-08-09
Fixed missing occupancy using local regression

mmmiah:
Oppps, I missed the pct of occupancy using local regression one. Just fixed it


ian-r-rose (approved):
LGTM, thanks @mmmiah

PR #366, opened 2024-08-23
fill in detector timestamps

### LINKED ISSUE #316: Fill in timestamp gaps in imputation
The imputation model is good at imputing data for timestamps that have bad or `null` data, but it currently doesn't fill in gaps if there is *no data reported at all*. So if a detector has a few hours where there are no data samples reported, and the timestamps don't exist in the raw data, then we won't impute for that gap. In practice, this does happen.

The task here is to fill in gaps in both stations and timestamps prior to imputing the data. This might be expensive, so doing it in a cost-effective way might be tricky. I would expect filling in station gaps to rely on some of the work @tnrahim is doing in #318.

JamesSLogan:
fixes #316

### DIFF
--- a/transform/macros/_macros.yml
+++ b/transform/macros/_macros.yml
@@ -103,3 +103,24 @@ macros:
           Each value for the partitioning string will be in the file path for the files,
           so it is a good idea to construct the expression with that in mind. For more documentation,
           see [here](https://docs.snowflake.com/en/sql-reference/sql/copy-into-location)
+
+  - name: timestamp_spine
+    description: |
+      Generates a table with 1 column; a list of timestamps from a start date to an end date. The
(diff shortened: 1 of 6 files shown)


kengodleskidot (approved):
Very nice work @JamesSLogan! The only comment I have is to keep @mmmiah in the loop since he is currently working on updating some imputation models and the `int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql` model is one he will want to be aware of.

JamesSLogan:
@mmmiah fyi - this PR introduces a new model (`int_clearinghouse__detector_agg_five_minutes_with_missing_rows`) that is being used as a drop-in replacement for `int_clearinghouse__detector_agg_five_minutes`. Models will now have more timestamps to impute over: 288 per detector per day.

PR #369, opened 2024-08-26
transfer aadt and k value table to mart for Dashboard Preparation

mmmiah:
We have calculated 8 different types of AADT and 3 different types of k value. However, this table is not available in mart for data visualization in GIS dashboard. We are working on PeMS visualization which needs aadt and k value. This PR just copying existing aadt and k value table  to mart for PeMS GIS virtualization dashboard.


mmmiah:
@kengodleskidot, could you please have your quick review for this PR. I am trying to use this table for one of the PeMS dashboard preparation. Thanks!

kengodleskidot (approved):
@mmmiah I updated the model name (misspelling) but other than that this is a simple, straightforward model.

Issue #370, opened 2024-08-27
Update Impuation Logic


mmmiah:
It seems that we have imputed the speed when volume and occupancy is zero. However, this speed should not be imputed rather it should be replaced with zero speed as there was no vehicle passed over the sensor in that time windows. Below is the screen shoot based on QC run by Ken-
![image](https://github.com/user-attachments/assets/3a92f133-c528-4a28-a9c3-e943b3786f9a)

Therefore, we need to update the imputation logic as

Volume =0 and occupancy=0, then speed=0

mmmiah:
completed

PR #371, opened 2024-08-27
Updated imputation logic (replaced null speed with zero when volume and occupancy are zero)

mmmiah:
This PR replaced null speed with zero speed when volume is zero and occupancy is zero to reflect correct imputation application. Previously we imputed null speed due to zeros volume and occupancy which mislead the imputation application. In this PR, we corrected those. This PR related to issues #370


mmmiah:
@kengodleskidot , please have your review on this!

kengodleskidot (approved):
Looks good, thank you @mmmiah! There is a merge conflict due to recent work done by James but you should be good to go once the conflicts are resolved.

mmmiah:
@JamesSLogan , my understanding is that I need to delete the old model {{ ref('int_clearinghouse__detector_agg_five_minutes') }} and keep yours, am I right?

image

Issue #379, opened 2024-08-29
Update models with normalized flow


jkarpen:
The `int_clearinghouse__detector_agg_five_minutes` model was previously updated to include both observed and normalized flow values which is detailed in #342. In order to ensure the downstream models use normalized flow values, the `int_clearinghouse__detector_agg_five_minutes` model will be updated so that the volume_sum variable is associated with the normalized value calculation instead of the observed flow value which can contain missing data. This will require a full data refresh of all downstream data models per my discussion with @JamesSLogan.

Issue #381, opened 2024-08-29
Add Imputation Method Values to Downstream Performance Models


kengodleskidot:
The imputation method values in the `int_imputation__detector_imputed_agg_five_minutes` model need to be added to the downstream 
`int_performance__detector_metrics_agg_five_minutes` model so they can be used in the reports and visualizations being developed. The imputation methods will also be used as selectable filters/parameters. There are three values that need to be added, which include: VOLUME_IMPUTATION_METHOD, SPEED_IMPUTATION_METHOD, and OCCUPANCY_IMPUTATION_METHOD.

PR #383, opened 2024-08-30
Update Normalized Flow and Imputation Methods

kengodleskidot:
The volume_sum field has been updated to use the normalized flow instead of the observed flow to account for any missing samples. I have also added the imputation method columns to the 5-minute aggregate detector performance model for use in reporting/visualization. This will require a full data refresh once the PR is merged.

### DIFF
--- a/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
+++ b/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
@@ -87,7 +87,9 @@ agg as (
                 sample_ct_{{ lane }} >= 10, volume_{{ lane }},
                 10 / nullifzero(sample_ct_{{ lane }}) * volume_{{ lane }}
             ))
-                as volume_sum, --Represents the normalized flow value
+                as volume_sum,
+                --Represents the observed or normalized flow value based on the
+                --number of samples recieved by the device
             zero_vol_ct_{{ lane }} as zero_vol_ct,
             occupancy_{{ lane }} as occupancy_avg,
             zero_occ_ct_{{ lane }} as zero_occ_ct,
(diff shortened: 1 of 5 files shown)


kengodleskidot:
This fixes #379 and #381

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes.sql
```diff
@@ -82,12 +82,12 @@ agg as (
...
             sample_ct_{{ lane }} as sample_ct,
             {{ lane }} as lane,
-            volume_{{ lane }} as volume_sum,
+            volume_{{ lane }} as volume_observed,
             round(iff(
                 sample_ct_{{ lane }} >= 10, volume_{{ lane }},
                 10 / nullifzero(sample_ct_{{ lane }}) * volume_{{ lane }}
             ))
-                as volume_normalized,
+                as volume_sum, --Represents the normalized flow value
```
JamesSLogan, line 90: After further review, I think this should be left as `volume_normalized` to better represent the column's meaning. This seems preferable to adding documentation+comments explaining what the column actually is doing. Of course, this means downstream models will need to be adjusted to use this new column name, which I think we wanted to avoid originally.

kengodleskidot: One of the issues with calling it volume_normalized is that it is not always a normalized volume value. If enough data points are reported by the device (which is usually the case), this value is simply the sum of all observed flow values. I believe is makes sense to update the comment to reflect this nuance and leave the variable name the same. What are your thoughts @JamesSLogan?

JamesSLogan: That sounds great - thanks!

JamesSLogan (approved):
Looks good

Issue #376, opened 2024-08-29
Implement alerting to know when different pipelines may be down


jkarpen:
From Ian:
There are internal dashboards at Caltrans for monitoring the data relay server and identifying any incidents or downtime. It would also be good to implement some freshness checks on the data consumer side to ensure that the raw data uploaded looks the way we expect.

One relatively straightforward way of doing this would be to implement a “[source freshness](https://docs.getdbt.com/docs/deploy/source-freshness)” check in our dbt project, which ensures that the source tables always have data that are at least as recent as some target date (often a few days in the past). We would create one source freshness check for each table involved in the data relay effort. This task would not need any special permissions in Snowflake, and could be implemented by anyone using the TRANSFORMER_DEV role.

Another approach could be to monitor the [copy history](https://docs.snowflake.com/en/user-guide/data-load-monitor) of the target tables for the data relay server to ensure that there are as many file uploads as we expect on any given date. This would require access to the SNOWFLAKE database, which would mean a bit of administrative work to ensure that the transformer roles can make the queries they need.

pingpingxiu-DOT-ca-gov:
The first approach is chosen.

jkarpen:
This is completed per @pingpingxiu-DOT-ca-gov

PR #390, opened 2024-09-05
apply source freshness check to data_relay sources

pingpingxiu-DOT-ca-gov:
The context is in this issue #376 
Goal: identifing outdated source tables, which indicates malfunctioning of data relay pipeline. 

Using dbt native function 
'''
dbt source freshness
'''

@ian-r-rose 
@jkarpen

### DIFF
--- a/transform/models/_sources.yml
+++ b/transform/models/_sources.yml
@@ -57,7 +57,7 @@ sources:
           Each line contains the sample timestamp and Station ID followed by flow,
           occupancy and speed for each lane. Note that occupancy and/or speed may be empty
           depending on the measurement capabilities of the detectors.
-        freshness: 
+        freshness:
           warn_after:
             count: 2
             period: day
@@ -169,7 +169,7 @@ sources:
     description: |
       Vehicle Detector Station (VDS) (aka Station) related data obtained via data relay from
       db96 within the Caltrans network.
-    freshness: 
+    freshness:
       warn_after:
         count: 2
         period: day
--- a/transform/models/_sources.yml
+++ b/transform/models/_sources.yml
@@ -57,6 +57,14 @@ sources:
           Each line contains the sample timestamp and Station ID followed by flow,
           occupancy and speed for each lane. Note that occupancy and/or speed may be empty
           depending on the measurement capabilities of the detectors.
+        freshness: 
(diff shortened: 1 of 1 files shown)


transform/models/_sources.yml
```diff
@@ -271,6 +288,7 @@ sources:
           stations hooked up to a single controller. For example, the same
           controller might have a station for the mainline as well as a station
           for an onramp.
+        freshness: null
```
kengodleskidot, line 291: @pingpingxiu-DOT-ca-gov where freshness is not needed for a table the example on dbt Developer Hub (https://docs.getdbt.com/reference/resource-properties/freshness) shows no argument after `freshness:` just a comment but you have included `null`. Did you run into any issues when executing this in the Dev environment?
image

kengodleskidot (approved):
Looks good @pingpingxiu-DOT-ca-gov I just had a minor comment regarding the code associated with tables that do not need a freshness check

transform/models/_sources.yml
peacexerox: Hi @kengodleskidot , this is good question, because in this case, adding or not adding null works, so it is styling. Styling is important for readability and maintenance. 
Rules of thumb: 
* If there is an existing style in place, then just follow it. Consistency of style is priority. 
* If this is the precedence, then I have my opinionated suggestion: adding an extra "null" gave us more keyword to search. So in order to search for the no-freshness-check occurances, we search for "freshness" and "null". This is easier than you first search for "freshness" and then eyeballing to further confirm this is "freshness: <empty>" rather than "freshness: ... <something>"

pingpingxiu-DOT-ca-gov: Hi @kengodleskidot , this is good question, because in this case, adding or not adding null works, so it is styling. Styling is important for readability and maintenance.
Rules of thumb:

If there is an existing style in place, then just follow it. Consistency of style is priority.
If this is the precedence, then I have my opinionated suggestion: adding an extra "null" gave us more keyword to search. So in order to search for the no-freshness-check occurances, we search for "freshness" and "null". This is easier than you first search for "freshness" and then eyeballing to further confirm this is "freshness: " rather than "freshness: ... "

PR #396, opened 2024-09-17
Update Station 5-minute Aggregation Model with Performance Metrics

### LINKED ISSUE #380: Update Performance 5-minute Agg Station Model with Performance Metrics
In order for delay values to be calculated for bottleneck and other station level analyses, performance metrics (VMT, VHD, Q, TTI, Delay and Productivity) need to be added to the `int_performance__station_metrics_agg_five_minutes` model. All other aggregates (detector and station level) models in the performance schema already contain these values so only this single table needs to be updated.

kengodleskidot:
Update max_capacity model to use a consistent value of 173 vehicles/lane/5-minute for each detector's max capacity. Also, update station 5-minute aggregation model with performance metrics. This fixes #380.

### DIFF
--- a/transform/models/intermediate/performance/int_performance__max_capacity.sql
+++ b/transform/models/intermediate/performance/int_performance__max_capacity.sql
@@ -29,7 +29,7 @@ max_capacity_detector as (
         */
         173 as max_capacity_5min
     from source
-    group by detector_id, station_id, physical_lanes
+    group by detector_id
 )
 
 select * from max_capacity_detector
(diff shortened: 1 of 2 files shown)


thehanggit (commented):
Thank you Ken! I saw the changes in station performance by including the delay metrics. Next week, I will create the bottleneck performance metrics to be consistent with the one in current PeMS system.

thehanggit (approved):

transform/models/intermediate/performance/int_performance__max_capacity.sql
```diff
@@ -1,40 +1,35 @@
...
+max_capacity_detector as (
+    select distinct
+        detector_id,
+        /*
+        2076 v/l/h / 12 = 173 v/l/5-min
+        */
+        173 as max_capacity_5min
     from source
-    qualify volume_sum is not null
+    group by detector_id
```
JamesSLogan, line 32: You can remove the group by since the aggregation was removed

JamesSLogan (approved):
Have you considered making this an ephemeral model? I'm not sure if there's much value in materializing a view in Snowflake here.

PR #401, opened 2024-09-25
Detector Status Model Update

kengodleskidot:
Updated detector_status and samples_per_detector models due to multiple Null values resulting in incorrect detector statuses. I believe the source was in how the metadata was being brought in.


mmmiah (approved):
Thanks for checking and updating! is there any change in upstream models that depends on this one???

kengodleskidot:
> Thanks for checking and updating! is there any change in upstream models that depends on this one???

The tables are nearly identical except that I added detector_type to the updated model.

Issue #386, opened 2024-09-04
Speeds and G-factors


thehanggit:
The PeMS system has the ability to compute speed for sensors that don't report speed, like single loop detectors. If the sensors are reporting speed, like with radar detectors deployed head-on, or double loop detectors, then we can use those speed measurement directly. This is a configuration option that needs to be specified when the system is configured. When we need to compute speeds, we follow the algorithm outlined below.

Traditionally speed has been computed by using a g-factor in combination with the flow and occupancy. The g-factor is a value that represents the effective length of the vehicle. It is a combination of the average length of the vehicles in the traffic stream and the tuning of the loop detector itself. Typically, a constant value for the g-factor is used which leads to inaccurate speeds because the g-factor varies by lane, time-of-day, as well as the loop sensitivity. PeMS estimates a g-factor for each loop for every 5 minutes over an average week to provide accurate speed estimates. The algorithm has been tested and validated against ground truth data from double loop detectors and floating cars.

The algorithm that we have implemented in the PeMS system is adapted from the paper ["Statistical methods for estimating speed using single-loop detectors,"](https://pems.dot.ca.gov/Papers/vanzwet_gfactor.pdf) by van Zwet, Chen, Jia and Kwon.

The steps for the estimation of speed from that paper are as follows:

They start by making the assumption that the speed on the freeway at free-flow conditions is known and constant. A couple of notes:
Free-flow is defined by having the occupancy less than a certain threshold.
The free-flow speed is only a function of the type of freeway (meaning the total number of lanes) and the particular lane that the detector is in. For example, the free-flow speed in lane 2 of a 2-lane freeway is different from the free-flow speed in lane 2 of a 5-lane freeway.
We have double loops in the Bay Area. We used these to measure the actual free-flow speeds on different types of freeways and in different lanes. We have a table that gives the free-flow speed for each lane of each type of freeway. We apply that to the entire state.
Using this assumption then for each loop we can work backwards and compute the g-factor for a number of points during a number of days.
We then smooth this using a robust adaptive regression method to obtain a g-factor for each loop in the system over a typical week. All of the computation for g-factors is done offline.
This g-factor is then used to compute the initial estimate of speed for each loop in real-time.
[...]

PR #406, opened 2024-09-26
enable udfs as macros

JamesSLogan:
1. Added `create_udfs` macro that gets called [on-run-start](https://docs.getdbt.com/reference/project-configs/on-run-start-on-run-end)
2. Added exponential smoothing udtf for use in #402 

Example udtf usage:
```sql
select
    detector_id,
    sample_timestamp,
    speed_preliminary,
    res.value_smoothed as speed_smoothed,
    p_factor
from
     int_clearinghouse__detector_g_factor_based_speed,
     TABLE(public.exponential_smooth(speed_preliminary, p_factor::float)
       OVER (PARTITION BY detector_id order by sample_timestamp)) AS res
```

### DIFF
--- a/transform/dbt_project.yml
+++ b/transform/dbt_project.yml
@@ -96,3 +96,6 @@ models:
       geo:
         +schema: geo
         +post-hook: "{{ unload_relation_as_geojson(strip_leading_words=1) }}"
+
+on-run-start:
+  - "{{create_udfs()}}"
--- a/transform/macros/create_udfs.sql
+++ b/transform/macros/create_udfs.sql
@@ -1,0 +1,5 @@
+{% macro create_udfs() %}
+
+    {{ create_exponential_smoothing_function() }};
+
+{% endmacro %}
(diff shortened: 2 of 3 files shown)


transform/dbt_project.yml
thehanggit, line 1: Question: Does that mean if I need to create any udfs, it is necessary to operate "on-run-start" in yml? Based on my understanding, it's like pre-defining the class/functions and let them be ready to be used anywhere, anytime in the project, right?

thehanggit (approved):
Thank you! James, I have completed the review and raised one trivial question. I will modify the model using udfs for recursive speed calculation.

transform/dbt_project.yml
JamesSLogan: `on-run-start` is automatic, so no manual operation will be required by a developer when building a model. The current setup will allow the UDFs to be used anywhere in the project 👍

thehanggit: Perfect!

ian-r-rose:
Cool stuff! I'm curious about the choice to use a python UDF rather than doing the computation in SQL. Is there something about the computation that makes a window function non-workable?

JamesSLogan:
> Cool stuff! I'm curious about the choice to use a python UDF rather than doing the computation in SQL. Is there something about the computation that makes a window function non-workable?

The function to be implemented (see #386) is recursive; each value relies on a previous value. A window function not being able to reference its own output is the main limitation (I'll be impressed if there is a way around this!). We did try a recursive CTE (see commits in #402) but hit some memory limitations we couldn't workaround.

PR #415, opened 2024-10-04
update incrementality of 2 intermediate models

### LINKED ISSUE #410: Debug Null Data in Clearinghouse 5-minute and Detector Status Models
@JamesSLogan after merging #408 yesterday I did some QC checks this morning and there are too many Null values in the `int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql` and `int_diagnostics__detector_status.sql` models for past dates. The last few days data looks good, but we have too many Null values for the older data. A full resfresh may fix this but we can discuss this in more detail at the sprint meeting this morning 10/3/24.

### LINKED ISSUE #414: Investigate performance regression
Since about two days ago, our nightly jobs went from taking about 30 minutes to about two hours. It seems that this is mostly being driven by [`int_clearinghouse__detector_agg_five_minutes_with_missing_rows`](https://california.getdbt.com/deploy/6/projects/10/runs/5449?tabId=model-timing), and the most likely culprit is something to do with #408 

@kengodleskidot and @mmmiah, do you know what might be going on there?

JamesSLogan:
#408 introduced new joins to these models that affected incremental builds. More specifically, the models are generating too many rows on incremental builds, which affects performance (#414) and overwrites rows with nulls (#410).

To test these changes, I ran these models a couple times after a full refresh to verify that they don't change.

I will execute a full-refresh of these models to fix existing production data issues when this PR merges.

Fixes #410 
Fixes #414

### DIFF
--- a/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
+++ b/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
@@ -15,6 +15,7 @@ with timestamp_spine as (
 
 detector_agg as (
     select * from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
+    -- This clause isn't strictly necessary but helps with performance on incremental builds
     where {{ make_model_incremental('sample_date') }}
 ),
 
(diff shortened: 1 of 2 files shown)


transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
ian-r-rose, line 1: I see that in #408 we also update the start date of the timestamp spine. Can we use the `pems_clearinghouse_start_date` variable instead of a hard-coded date here? Just looking forward to when we roll that start date back: it will be good to have to change it in as few places as possible.

ian-r-rose (approved):
Thanks @JamesSLogan this looks good to me. I have just one (optional) comment

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
JamesSLogan: Done 👌

kengodleskidot:
Thank you for making these updates @JamesSLogan. When I tried adding the incremental macros in the CTE's during my PR I ran into errors, so I am glad you were able to get these updated so quickly!

kengodleskidot (approved):
Thank you @JamesSLogan

Issue #413, opened 2024-10-03
Meta Issue: List of PeMS Data QC visulization Task


mmmiah:
We have created several datasets along with QC data table. However, it is difficult to understand how this QC looks like. Therefore, it is necessary to monitor the data QC through visualization. I am proposing below list of the QC to ensure the high quality data production in PeMS. 

1.  A dashboard showing number of daily detectors side by side (may be bar chart) from the metadata, five minutes detector models with and without imputation. The bar value can be absolute number or percentage of difference between the number of detectors. This dashboard can be build dynamically based on last seven days data from current date which will provide temporal data quality assurance over time. This dashboard visualization can be further break down by district, route, county and others meta data catagory to bring spatial data quality understand. Issue #425
2.  Similar to Detector level QC dashboard, station level daily number of detectors for last seven dynamic days can be visualized to ensure data quality in station level. issue #427
3.  Compare the Number of good and Bad (details Category) detectors reported in  
INT_DIAGNOSTICS__REAL_DETECTOR_STATUS over consecutive time. For example, show the number of detectors for each error type over seven days to understand any unusual detector status. issues #428
4.  Only comparing number of detectors or stations over spatial and temporal period does not provide 100% guaranty of good data quality. To ensure good data quality we need to have long term visualize plan of some of the metrics spatially and temporally to understand their pattern and outliers. The below is the subcategory for some of those visualization-
   a. HOV Vs mainline Performance evaluation in terms of volume, speed and occupancy for a one week recent time series. issue #429
   b.  Detector/Station level imputed Vs. observed volume, occupancy, VMT and VHT line diagram for a week to understand any unusual observed or imputed value. issue #430
   c.  Detector/Station Level AADT for last 2/3 years to check if there is abnormal drop of AADT value between two/three consecutive years for a same station or detector. This AADT check will cover the others micro level QC between annual and daily level. If the Bigger picture looks good, then other small pictures should be fine. issue #431
d. Compare the 5mins weekday and weekend volume, speed and occupancy between old PeMS and New PeMS #464 

I will add more in later and feel free to add any other ideas of QC visualization if you can remember.

jkarpen:
@mmmiah will create separate issues for these tasks for tracking purposes.

mmmiah:
All of these listed QC and reporting dashboard was completed

Issue #398, opened 2024-09-19
Diagnostic Model QC


kengodleskidot:
As part of our data quality efforts related to detector health, we will compare the daily number of detectors in the existing PeMS detector_status table data for existing PeMS to the `int_diagnostics__detector_status` data model for this project for consistency. We will also compare counts for the daily detector status results to compare similarities/differences in the number of good/bad detectors as well as the errors associated with bad detectors.

kengodleskidot:
related to #497

Issue #397, opened 2024-09-19
Clearinghouse Model QC


kengodleskidot:
In the clearinghouse schema we have the `int_clearinghouse__detector_agg_five_minutes_with_missing_rows` data model which contains an aggregation of all observed data in five-minute increments as well as rows for 5-minute timeframes where we do not have observed data. Rows are generated for all 288 5-minute timeframes in a day per active detector (12 5-minute time periods/hour * 24 hours/day). For 5-minute timeframes where we don't have observed data, null values represent the missing flow, occupancy, speed, sample count and other data aggregations but the detector/station metadata and date/timestamp data will be populated. Any missing data will be filled in with imputed data as part of the downstream data models built in the imputation schema. 

The QC goal for the clearinghouse model is to ensure that we have all active detectors reporting 288 rows on a daily basis that meet with test requirements contained in the associated yml files (e.g. detector_id, freeway, direction and other rows that should have data do not return null values). The daily detector counts between the `int_diagnostics__detector_status` and `int_clearinghouse__detector_agg_five_minutes_with_missing_rows ` models should be consistent. In addition, other downstream models will be compared for consistency against the aforementioned models as well.

PR #416, opened 2024-10-04
Data Quality Daily Row Check Mart Model

kengodleskidot:
The data model created in this PR can be used to create data quality visualizations/reports. The model aggregates the daily number of detectors for 4 data models that should all match. Below is an example of how two of the models are matching but subsequent models do not along with a screenshot in PeMS for comparison purposes:
image

image

I anticipate the Good and Bad counts to update as needed after the full data refresh. This should assist with #413, #397 and #398.

### DIFF
--- a/transform/dbt_project.yml
+++ b/transform/dbt_project.yml
@@ -101,6 +101,8 @@ models:
       geo:
         +schema: geo
         +post-hook: "{{ unload_relation_as_geojson(strip_leading_words=1) }}"
+      quality:
+        +schema: quality
 
 on-run-start:
   - "{{create_udfs()}}"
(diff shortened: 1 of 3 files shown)


kengodleskidot:
> The data model created in this PR can be used to create data quality visualizations/reports. The model aggregates the daily number of detectors for 4 data models that should all match. Below is an example of how two of the models are matching but subsequent models do not along with a screenshot in PeMS for comparison purposes: image
> 
> image
> 
> I anticipate the Good and Bad counts to update as needed after the full data refresh. This should assist with #413, #397 and #398.

@mmmiah pointed out that we should only be comparing detector counts for station types of ML and HV only since the the imputed and performance data models are for those station types only. The mart data model has been updated to reflect these station types. The updated model reflects row counts that match, great catch @mmmiah!

image

We will still include the good/bad/total detector count from the detector_status model for QC usage.

mmmiah (approved):
Please merge! I will use this data for QC visualization from analytical_prd

PR #417, opened 2024-10-04
change unload sql to save as csv instead of parquet format

ZhenyuZhu-Caltrans:


jkarpen:
Per @ZhenyuZhu-Caltrans keeping this in the backlog until further discussion is done on how to proceed.

PR #419, opened 2024-10-09
Ten days corridor level performance mart model bug fixing

mmmiah:
There was bugs in logic that generate just 10 days back single day data, while our target was getting recent all 10 days data. it seems the data size is too big, I am limiting this data for just a single recent week

### DIFF
--- a/transform/models/marts/performance/_performance.yml
+++ b/transform/models/marts/performance/_performance.yml
@@ -346,3 +346,70 @@ models:
           The lost Productivity perfromance metric is
           the number of lane-mile-month on the freeway lost due to operating under congested
           conditions instead of under free-flow conditions.
+  - name: performance__station_metric_agg_recent_one_week
+    description: |
+      Hourly aggregation of volume, vmt, and vht for HOV and mainline pairs for recent one week.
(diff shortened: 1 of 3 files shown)


mmmiah:
@kengodleskidot , this is ready for re-review my edits

kengodleskidot (commented):
On line 133 I believe you included vht into a calculation that should use vmt, see screenshot below. Also is there a yaml file or comments you can include about this model  that describe what the penetration values represent and the level of comparison being made? It looks like an average per lane comparison based on the calculations. A description of what acceptable values would be (possibly a range) or other context would help others in understanding what this analysis represents.

image

mmmiah:
> On line 133 I believe you included vht into a calculation that should use vmt, see screenshot below. Also is there a yaml file or comments you can include about this model that describe what the penetration values represent and the level of comparison being made? It looks like an average per lane comparison based on the calculations. A description of what acceptable values would be (possibly a range) or other context would help others in understanding what this analysis represents.
> 
> image

Great catch! I just fixed it and also added yml file!

kengodleskidot (approved):
@mmmiah looks good to go!

mmmiah:
closed as completed

Issue #431, opened 2024-10-15
AADT Pattern Monitoring by corridor over multiple years


mmmiah:
As part of the QC, we will utilize our mart model named 'PERFORMANCE.PERFORMANCE__STATION_AADT_K_VALUE' that have reported historical AADT and K -factors for each station. We can easily use this data table to grab last 2-3 years AADT data and show some visualization by corridor or freeway/route level. Over the consecutive time period AADT should not drop too much by the corridor level. If there is unusual drop of AADT noticed, we need to investigate our data processing. This dashboard will be dynamic using last 2-3 years data and update the map yearly basis using scheduler of corn job in logstash and visualization will be done in kibana through elastic search data absorption.

mmmiah:
Here is the aadt kibana dashboard draft for the review: Please have your review and let me know your feedback. Thanks!
https://svgcmdl02:5601/app/dashboards#/view/66a3c070-90b6-11ef-af1a-8553ba48f88d?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2y,to:now))&_a=(description:'',filters:!(),fullScreenMode:!t,options:(hidePanelTitles:!f,useMargins:!t),query:(language:kuery,query:''),timeRestore:!f,title:station_aadt_k-value_performance_dashboard,viewMode:view)

thehanggit:
Thank you Mintu for your great efforts! The AADT by freeway is bit unclear and the filter window is somehow blocked by the map. Can you please adjust them?

thehanggit:
> Thank you Mintu for your great efforts! The AADT by freeway is bit unclear and the filter window is somehow blocked by the map. Can you please adjust them?

By "unclear" I mean it's hard to differentiate different AADTs

mmmiah:
@thehanggit ,Thanks for quick review. Do you see different than me
![image](https://github.com/user-attachments/assets/4717bbeb-d08a-4a0c-a51c-5bd7289fa547)
 We calculated eight different AADT, those are looks have minor difference from each others due to their calculation criteria. As long as those are close to each others our calculated AADT looks good!

thehanggit:
Yes, it's the same dashboard. Looks good!

jkarpen:
Per @mmmiah this dashboard is completed and has received feedback so closing this issue.

PR #433, opened 2024-10-16
Included county and lat-long in aadt model

mmmiah:
This PR related to QC visualization issues #431. Currently in our existing PERFORMANCE__STATION_AADT_K_VALUE we  do not have county full name and also lat and long is missing for the station. Doing inner join within logstash between two data frame is time consuming and costly to have lat-long. shows memory error. I tested following query 
image
 it stucked the query layer within logstash that indicates that we need mostly ready to feed data from mart to elastic search for the visualization. For that need, I am adding county and lat_long within dbt environment and sending this edited model to mart, so that I can directly read and push the data in elastic search instead of doing query layer within logstash which is burden.


JamesSLogan (approved):
Looks good to me, thanks for adding the documentation. I'm not sure if it'll help Kibana at all, but you could consider [converting](https://docs.snowflake.com/en/sql-reference/functions/st_makepoint) the lat/long values to a geography type (like in the `geo__current_detectors` model).

mmmiah:
@JamesSLogan , Thanks James, Logstash can convert lat and long to co-ordinates using below code:
image
However, I realize that we can convert lat and long to location string (array was preferable but pop up error) which can be directly read by elastic search as geo-point.

mmmiah:
close as completed

Issue #430, opened 2024-10-15
Create a daily 24 hours dynamic visulization of imputed vs observed metrics (volume, speed, occupancy, VMT,  and VHT)


mmmiah:
We used five different imputation techniques which includes local, regional, global regression and local/regional average to impute the missing data. However, it is necessary to monitor how the imputed values are aligned with observed values over 24 hours daily basis. We will develop similar daily level dynamic graphs that Ian already demonstrated in python script in this page: https://cagov.github.io/caldata-mdsa-caltrans-pems/data/imputation/. We will do something similar in daily level ETL in kibana dashboard using logstash while read the data from snowflake mart.

PR #434, opened 2024-10-16
create a copy of imputed 5 mins data to mart for visualization purpose

mmmiah:
I have copied 5 mins imputation table to mart for the visualization purposes based on issue #430. This dataset in mart level was limited for recent 3 days 5 mins imputed plus observed data for HOV and ML only


ian-r-rose:
This will run into trouble without some partitioning, as the automatic unloading to S3 cannot upload a ~300 GB table in one file. I think if you want to have large tables in the marts database, we'll need to do some more specific planning for it (including some schemas that only create views and don't unload the whole table to S3)

ian-r-rose:
Your approach to limiting the data also works! But if you want the full table, there are some things we might explore.

mmmiah:
Agree, limited the data for recent 3 days for HOV and ML only, now the data size is 10.8 M!

kengodleskidot (approved):
Waiting for the PR check to pass but this one is straight forward.

mmmiah:
3 days data is almost 30 M which exceed the max size 512 MB, reduce the data just for recent one day which is 10.8 M and pass the 512 MB size limit! Hopefully now the dbt check will pass!

ian-r-rose:
@mmmiah I recommend using partitioning to write multiple files. This allows you to write larger datasets to S3.

ian-r-rose:
If you want to grab 30 minutes on my [calendar](https://calendar.google.com/calendar/u/0/appointments/schedules/AcZssZ1DvKZPVPUEDE9O3yI8gJXQUfMiYVaTg0_HUUrnRPPHp1Si-XuEK_GaqTj62T8AFzidRZtES14d) I can show you how to do it

mmmiah:
> If you want to grab 30 minutes on my [calendar](https://calendar.google.com/calendar/u/0/appointments/schedules/AcZssZ1DvKZPVPUEDE9O3yI8gJXQUfMiYVaTg0_HUUrnRPPHp1Si-XuEK_GaqTj62T8AFzidRZtES14d) I can show you how to do it

Let me catch you in our regular meeting for 10-20 mins at the end of meeting to learn more! Partition concept is interesting. I have just applied the below partition for recent 3 days 30 M data and it works really fine. Thank you for heads up!

{{ config(
    materialized="table",
    unload_partitioning="('day=' || to_varchar(date_part(day, sample_date)) || '/district=' || district)",
) }}

PR #436, opened 2024-10-21
Make the detector_config_log freshness check less sensitive

pingpingxiu-DOT-ca-gov:
Adjusted the freshness warning to 14 days

### DIFF
--- a/transform/models/_sources.yml
+++ b/transform/models/_sources.yml
@@ -169,14 +169,6 @@ sources:
     description: |
       Vehicle Detector Station (VDS) (aka Station) related data obtained via data relay from
       db96 within the Caltrans network.
-    freshness:
-      warn_after:
-        count: 2
-        period: day
-      error_after:
-        count: 3
-        period: day
-    loaded_at_field: time_id
     tables:
       - name: vds30sec
         description: |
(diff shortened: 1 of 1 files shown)


kengodleskidot:
> Adjusted the freshness warning to 14 days

This adjustment seems reasonable based on the frequency that districts update the equipment set files and now that the district 8 automated update for the equipment set file is not being done daily.

ian-r-rose (approved):
Does it also make sense to do this for the station and controller config logs?

ian-r-rose:
@pingpingxiu-DOT-ca-gov just wanted to make sure you saw my question above

pingpingxiu-DOT-ca-gov:
Per our discussion, station/controller changing frequencies are very different on detectors. 
So the scope was on detector. @ian-r-rose

transform/models/_sources.yml
```diff
@@ -169,14 +169,6 @@ sources:
     description: |
       Vehicle Detector Station (VDS) (aka Station) related data obtained via data relay from
       db96 within the Caltrans network.
-    freshness:
```
ian-r-rose, line 172: It didn't really make sense to have this config at the schema level, as not every table has a `time_id` column. Removing this allows us to remove the `null` overrides below, and we can just add more specific tests on the tables we actually want to check

ian-r-rose:
This is now ready for final review, and should make our freshness check stop tripping!

JamesSLogan (approved):
Yay!

Issue #437, opened 2024-10-21
Daily Imputed and Non imputed sample count by Imputation Methods


mmmiah:
We have developed the imputation model, previously we created the station level summary to understand what percentage of sample each day in each station are imputing by imputation methods. Similar to previous mart model, we want to monitor the daily sample size from imputation model and want to break down how many samples were imputed by different methods. This aggregation will be by sample date, regardless of station/detector to understand the granular level data QC.

PR #438, opened 2024-10-21
developed imputation QC model that will monitor daily sample balance for different imputation methods

mmmiah:
This PR is built based on issue #437 which counted the daily sample size by different imputation and check the balance of sample size between the daily observed and imputation for QC purpose


transform/models/marts/quality/quality_imputation_daily_sample_count.sql
```diff
@@ -0,0 +1,83 @@
...
+        (vol_unobserved_unimputed / nullif(sample_ct, 0)) * 100 as pct_vol_unobserved,
+        (speed_imputed_sample / nullif(sample_ct, 0)) * 100 as pct_speed_imputed,
+        (speed_observed_sample / nullif(sample_ct, 0)) * 100 as pct_speed_observed,
+        (speed_unobserved_unimputed / nullif(sample_ct, 0)) * 100 as pct_speed_unobserved,
+        (occ_imputed_sample / nullif(sample_ct, 0)) * 100 as pct_occ_imputed,
+        (occ_observed_sample / nullif(sample_ct, 0)) * 100 as pct_occ_observed,
+        (occ_unobserved_unimputed / nullif(sample_ct, 0)) * 100 as pct_occ_unobserved,
+
+        -- Volume check: Sum of all volume percentages should equal 100
+        coalesce(
```
JamesSLogan, line 61: Do you think adding these checks as tests instead of columns would be a better way of raising issues? That way if somebody in the future, for example, adds a new imputation method not captured here, the test(s) would automatically flag that this model needs changes as well.

ian-r-rose: I'm not sure that adding these as columns is the right way to go, it may be better to add them as tests *only*.

In general, it is very unreliable to do floating point math and expect strict equality checks to work. This is due to fundamental limitations in how floating point numbers are represented internally. [Here](https://docs.snowflake.com/en/sql-reference/data-types-numeric#rounding-errors) is some recommendation from snowflake on the matter. The usual way to write this kind of test is to instead assert that the absolute value of the difference of the numbers is less than some small number.

ian-r-rose (commented):
I agree with @JamesSLogan that moving the checks into tests would be more appropriate. I also recommend using a more precision-safe check for those checks.

transform/models/marts/quality/quality_imputation_daily_sample_count.sql
mmmiah: I agree to drop those numerical test within the data frame. I have dropped all numerical sum test and will add separate yml test for that may be later

mmmiah:
@kengodleskidot , would you be able to add three test on this imputation summary table in your previous test file. One for speed, one for volume and another for occupancy. Observed, imputed and observed_unimputed sum for each metric should be 100 or super minor less
image

transform/models/marts/quality/_quality.yml
```diff
@@ -34,3 +34,79 @@ models:
           Counts the total number of detectors with a status other than Good on a daily basis.
       - name: ALL_DETECTOR_STATUS_TOTAL_COUNT
         description: Counts the total number of rows/detectors on a daily basis
+  - name: quality_imputation_daily_sample_count
+    description: |
+      This model counts the number of imputed, observed and observed-unimputed daily sample as
+      well as their percentage. The count of imputed sampe was again breakdon by imputation methodology.
```
JamesSLogan, line 40: sampe -> sample
breakdon -> broken down

mmmiah: opps! fixed the typos!

mmmiah:
@JamesSLogan , Could you please look on this PR whenever you get time. It is failed due to dbt CI failure of 00:24:39 11 of 83 ERROR creating sql incremental model dbt_cloud_pr_15_438_diagnostics.int_diagnostics__constant_occupancy  [ERROR in 1.70s]. I tried to drop the relevant one, but still failed. Thanks!
image

summer-mothwood:
@JamesSLogan this error in the constant occupancy code is happening because Mintu is working off a version of the code before the change I applied to remove station id from its output. I'm assuming the best way to address this is to rebase Mintu's branch to the current version of the code then trying again. Is there a good way to do that in dbt Cloud?

ian-r-rose:
I learned the hard way that attempting to rebase with dbt Cloud is a recipe for heartache. If using cloud, it better to just merge from `main`.

mmmiah:
Woow, It seems that it passed the dbt failure! If you do not have any comments, please go ahead and approve this PR, so that we can merge it to main!

JamesSLogan (approved):
Thanks Mintu! Has an issue been created for adding the tests discussed in this PR?

mmmiah:
> Thanks Mintu! Has an issue been created for adding the tests discussed in this PR?

Ken is supposed to add test separately. No test in this PR

PR #440, opened 2024-10-22
Fix data status incrementality

### LINKED ISSUE #426: Detector Status Not Updating Properly
I reviewed @mmmiah screenshot in #425 and checked the detector_status model data in Production and it appears the data in the Production detector_status model is not updating properly. Below and attached are the results when I build the model in Dev which look correct. We should see approximately 25k Good detectors on any given date but in Production the statuses are all bad since 10/8/24. 
[PRD-Diagnostic.csv](https://github.com/user-attachments/files/17385070/PRD-Diagnostic.csv)
image

JamesSLogan:
There have been some recent changes to have models use the active station/detector models as a starting point, then left join in "source" data. This is helpful for consistency but also currently creates an issue with incremental models. Why: the "active" models use vds metadata as a starting point, which is a day or two more recent the "source" clearinghouse data. This slight discrepancy means that some incremental builds will look farther ahead than the source data contains (this is due to the left join mentioned above). This PR removes the assumption of identical starting dates from 2 models by limiting dates when selecting from "active" models.

This will have an impact on models downstream from the detector status model, in that they might appear to "lose" ~2 days worth of not-useful data. 

Fixes #426


kengodleskidot (approved):
Thank you for your work on this and finding the source of the issue @JamesSLogan! Did you come across a noticeable increase in the build time? I'll perform a review of the data after the next update and include any feedback.

PR #445, opened 2024-10-24
Update unique keys to allow incremental builds

### LINKED ISSUE #444: Investigate recent CI failure
Ask is to investigate recent CI failure, more details here: https://california.getdbt.com/deploy/6/projects/10/runs/5842

JamesSLogan:
Fixes #444 

I tested that incremental builds do succeed with this change. Downstream models don't appear to need changes.

### DIFF
--- a/transform/models/intermediate/performance/int_performance__bottleneck_delay_metrics_agg_daily.sql
+++ b/transform/models/intermediate/performance/int_performance__bottleneck_delay_metrics_agg_daily.sql
@@ -1,6 +1,6 @@
 {{ config(
     materialized="incremental",
-    unique_key=['station_id','sample_date'],
+    unique_key=['station_id','sample_date', 'time_shift'],
     snowflake_warehouse = get_snowflake_refresh_warehouse(small="XL")
 ) }}
 


thehanggit (approved):
Thank you @JamesSLogan! Do you know why it is necessary to have time shift as unique keys? I'm curious if I need to group by any variable in the future, should I also consider that variable into unique keys when implementing the incremental model.

JamesSLogan:
We use [unique keys ](https://docs.getdbt.com/docs/build/incremental-models#defining-a-unique-key-optional) so that dbt doesn't simply append new rows on each incremental build. The unique keys do need to uniquely identify a row in the model when they are used. In models where the final query includes a `group by`, the grouping columns are definitely a good clue as to what columns to specify for unique keys! 👍

thehanggit:
> We use [unique keys ](https://docs.getdbt.com/docs/build/incremental-models#defining-a-unique-key-optional) so that dbt doesn't simply append new rows on each incremental build. The unique keys do need to uniquely identify a row in the model when they are used. In models where the final query includes a `group by`, the grouping columns are definitely a good clue as to what columns to specify for unique keys! 👍

Great! That means there are rows having same station ids and dates but with different time shift so they can not build and that's why we need to add it into unique keys. Thanks for knowledge sharing!

ian-r-rose:
Thanks for the quick fix!

Issue #448, opened 2024-10-28
Duplicate data in aadt and K-factor model


mmmiah:
Recently I developed the dashboard for monitoring aadt and k factor. While working on dashboard update, I noticed that duplicate data generated in the snowflake. Lets investigate the root of this duplicate and fix the problem. Example of duplicate data 
![image](https://github.com/user-attachments/assets/cf0b9b8a-7355-419a-8488-b993f13813b3)
![image](https://github.com/user-attachments/assets/e4e0b5fd-3da6-4539-9835-02d388574af8)

PR #450, opened 2024-10-28
Bottleneck Dashboard

thehanggit:
PR related to #448


transform/models/marts/performance/performance__bottleneck_delay_agg_daily.sql
mmmiah, line 1: `county as (
    select
        county_id,
        county_name
    from {{ ref('counties') }}
),

bottleneck_delay_with_county as (
    select
        daily_bottleneck_delay.* exclude (county),
        c.county_name
    from
        daily_bottleneck_delay
    inner join
        county as c
        on daily_bottleneck_delay.county = c.county_id
),`

replace above code with 
`bottleneck_delay_with_county as (
    {{ get_county_name('daily_bottleneck_delay') }}
)`

thehanggit: Thanks Mintu! I already updated it.

mmmiah (approved):
looks good! Merge when all CI pass

PR #452, opened 2024-10-29
replace the valid_from with active_date to have recent meta data for PERFORMANCE__STATION_METRIC_AGG_RECENT_ONE_WEEK model

mmmiah:
I have replaced the valid from with active date from {{ ref('int_vds__active_stations') }} to have recent active station. We found that detectors under a station may be located in adjacent city and county. The QC link is here: https://app.snowflake.com/vsb79059/dse_caltrans_pems/w5IeSThdLHF7#query


kengodleskidot (approved):
Looks good to me @mmmiah and the QC checks passed when I ran them based on the link you have in the PR. Let me know if you need anything else.

PR #456, opened 2024-10-31
Clearinghouse Models QC Checks

### LINKED ISSUE #397: Clearinghouse Model QC

kengodleskidot:
Added data QC checks for the clearinghouse models in the yaml file and created a custom dbt check modeled after the demo @JamesSLogan provided. I also cross referenced the tables/yaml files to try and capture additional data field descriptions and update as needed. This fixes #397.


mmmiah (approved):
looks good! Please check the CI test failure

transform/tests/generic/five_minute_daily_count.sql
```diff
@@ -0,0 +1,12 @@
+{% test five_minute_daily_count(model, select_columns, group_by_columns) %}
```
JamesSLogan, line 1: Minor suggestion: I think I'm in favor of removing `select_columns` and substituting `group_by_columns` in the select below. The `select_columns` only exists because select * is invalid and I don't currently see an instance where we would select something we haven't grouped by. Sorry for the bad example code 😆

kengodleskidot: @JamesSLogan I reviewed the CI failure in the screenshots below and replaced the single quotes with double quotes in a previous commit but this resulted in more failures. I believe the existing formatting is correct, is there a way someone from ODI can review this check and confirm the error is valid? I'll address your comment above as well 👍
image
image

JamesSLogan: You can get around this by swapping your single and double quotes here. Your syntax is correct but the project currently requests double quotes in yaml files...for consistency. [Config can be changed here](https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/main/.yamllint)

kengodleskidot: I swapped single for double quotes and it failed even more checks which is why I swapped them back to single quotes.
image

JamesSLogan: Sorry for the confusion! If you make the outer quotes double quotes and the inner quotes single, it should solve the issue. It looks like the reverted commit made them all double.

`values: "{{ var('districts') }}"`

Issue #457, opened 2024-11-01
Add New Tag to imputation Model


mmmiah:
There are several case where none of the imputation method was eligible to apply to fill the observe null values . Previously we tag those as observed. However, in our meeting we decided to tag this kind of case as "Observed_unimputed"  to avoid any confusion by the tagging

mmmiah:
this task is completed

PR #458, opened 2024-11-01
added new tag as 'observed_unimputed" when null metric was not impuited

mmmiah:
This PR related to issue #457. We added a new tag for the observed unimputed null value.


ian-r-rose (approved):
Looks good, thanks @mmmiah

PR #469, opened 2024-11-08
make clearinghouse load sooner

### LINKED ISSUE #455: Make data available just one day lag from current date in full refresh
Right now in our production there are at least 3 days lags of data in different models which is fill with null values. When we will do full refresh, please make sure that we have match different model production just one day lag from current date, so that all models output are current in real time up to yesterday. 
image

JamesSLogan:
Fixes #455 
^mostly fixes, at least. This change puts the s3 loading 2 hours prior to our nightly build, which will decrease daily lag by 1 day. Specifically, lag will go from 3 to 2 days. To get us to 1 day, we would need to load data as soon as it becomes available in clearinghouse.

Currently, "yesterday's" data arrives during the hours of 4:00 to 5:30 AM the next day. To schedule the current process of loading to s3, we would want to start it ~6-7 to be safe. This would lead to kicking off the nightly build around 7-8 AM, which is arguably too late to guarantee data availability before people start work, some days, at least. I think it's worth it to keep the current implementation, especially considering that the data relay server should improve this latency in the future.

@mmmiah fyi

### DIFF
--- a/dags/load_from_clearinghouse.py
+++ b/dags/load_from_clearinghouse.py
@@ -99,14 +99,13 @@ def clearinghouse_to_s3(day: date) -> None:
 
 @task
 def pems_clearinghouse_to_s3_task(**context):
-    prev_date = (context["execution_date"] - timedelta(days=1)).date()
-    clearinghouse_to_s3(prev_date)
+    clearinghouse_to_s3(context["execution_date"].date())
 
 
 @dag(
     description="Load data from the PeMS clearinghouse webserver to S3",
     start_date=datetime(1994, 12, 1),
-    schedule_interval="0 8 * * *",  # 12 AM PST
+    schedule_interval="0 14 * * *",  # 6 AM PST
     default_args={
         "owner": "Traffic Operations",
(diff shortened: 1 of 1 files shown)


JamesSLogan:
Thanks for the feedback @ian-r-rose , it is now implemented. I believe the following manual actions will need to take place post-merge:
1. Edit nightly build schedule (I can do this)
2. Deploy dag to airflow (Ian?)
3. Re-run dag for "yesterday" since we will be skipping a day when this is merged/deployed (Ian?)

ian-r-rose:
Update here @JamesSLogan, this is now deployed to Airflow, and we are now caught up (data from this morning is in Snowflake!)

Also, I made a mistake above. This script doesn't take 60 _minutes_, it takes 60 *seconds*. So actually I think it would be quite safe to schedule the "nightly" job for 6:30 AM.

JamesSLogan:
Awesome, thank you! I re-updated the dbt job for 6:30. 🤞 for tomorrow's run

PR #471, opened 2024-11-08
change constant_occupancy model schema to detector_id

summer-mothwood:
Changed one of the diagnostic models to output and merge on detector_id rather than the surrogate station_id + lane combo. This necessitated a downstream code change and a documentation change as well.


summer-mothwood:
this change will require a full refresh of the constant_occupancy model once merged. 

I don't think the downstream models will need to be fully-refreshed as well, because the changed columns were only used in a join condition and there shouldn't be any schema changes or other major issues as a result once merged.

ian-r-rose (approved):
Looks good, thanks!

PR #474, opened 2024-11-12
change unique key to detector_id for constant occupancy model

summer-mothwood:
trying this again! same code changes as PR #471 , but this time I'll run the full-refresh of constant occupancy and its downstream models in production after the code is merged so that the nightly build doesn't fail. :)


ian-r-rose (approved):
Looks good

PR #476, opened 2024-11-13
QC Updates for Diagnostic yaml File and Detector Status Model

kengodleskidot:
The yaml file has been updated with missing documentation and tests have been added. I have also removed the Station Down status since it does not currently exist in the current PeMS and feedback from district is needed to see if it provides any value. Once this PR is ready for merging the `int_diagnostics__detector_status` will need a full refresh. Linking #398 since this is related.

### DIFF
--- a/transform/models/intermediate/diagnostics/_diagnostics.yml
+++ b/transform/models/intermediate/diagnostics/_diagnostics.yml
@@ -289,18 +289,14 @@ models:
       hours) where occupancy is not null and not zero. Finally it outputs the
       minimum value grouped by detector_id and sample_date.
     columns:
-      - name: station_id
+      - name: detector_id
         description: |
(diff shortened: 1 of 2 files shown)


mmmiah (approved):
Looks great to me! 
Minor comment (optional to change):
`when nds.detector_feed_working = 'No' then 'No Data'`
does this make sense to say ' `when nds.detector_feed_working = 'No' then 'Detector Down'`
like others.

kengodleskidot:
> Looks great to me! Minor comment (optional to change): `when nds.detector_feed_working = 'No' then 'No Data'` does this make sense to say ' `when nds.detector_feed_working = 'No' then 'Detector Down'` like others.

I used "No Data" to be consistent with what the current system uses. If the team agrees that "Detector Down" is a better description than I would concur with you and change it:
image

PR #481, opened 2024-11-19
swap out station id + lane for detector id as unique key in imputation models

summer-mothwood:
Edited the imputation models to work off of detector id. Two notes:
1) This will require a full refresh after the code is merged to account for detector id now being an output column in several of these models. I plan to do this in 2 phases -- first a full refresh of the global coefficient model on its own, then a full refresh + downstream of the local/regional coefficient model. This approach limits the number of times the same downstream models are refreshed, since both coefficient models feed into the same downstreams.

2) we still use station ID to find the nearby detectors in the local and regional models. do we want to create a 'nearby detectors' model as well? I don't think it's necessary, unless detectors ever move to other stations, but could use a second opinion.


mmmiah (approved):
@summer-mothwood , looks good to me! Thanks for doing these, it seems that you compacted the model very well and saved aggregation time as well!

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
```diff
@@ -88,14 +88,13 @@ unimputed as (
         -- If the station_id in the join is not null, it means that the detector
```
ian-r-rose, line 88: This comment is out of date

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
```diff
@@ -195,8 +194,7 @@ samples_requiring_imputation_with_local_regional_coeffs as (
     asof join local_regional_coeffs
         match_condition (samples.sample_date >= local_regional_coeffs.regression_date)
         on
-            samples.station_id = local_regional_coeffs.station_id
-            and samples.lane = local_regional_coeffs.lane
+            samples.detector_id = local_regional_coeffs.detector_id --change this when done testing
```
ian-r-rose, line 197: What needs to be changed here when done with testing? This looks like a correct join condition to me

ian-r-rose (commented):
Thanks @summer-mothwood, all of your join condition changes look good to me.

I see that you've also removed `lane` from most of the models. I wonder if it would be worth keeping it around as a useful piece of denormalized metadata. I could go either way, but maybe @mmmiah could comment?

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
mmmiah: Keeping lane with detector id in final data will be obviously much useful! We need lane number for lane to lane analysis. for example, most inner lane performance or most outer lane performance. Also for D4, lane 1 is the HOV lane, without lane number, it is difficult to identify with detector id only. I will say please have lane number in the final data, although we can retrieve lane number using detector conf, but that needs additional effort to bring two data together.

summer-mothwood: Great catch @ian-r-rose and thank you for the detailed context @mmmiah ! I'll add lane back in.

transform/models/intermediate/imputation/int_imputation__detector_agg_five_minutes.sql
summer-mothwood: an out of date comment to myself during development! thanks for catching.

PR #487, opened 2024-11-26
update to latest major version of dbt-core and dbt-snowflake

summer-mothwood:
Need to be on 1.8 to set/test the new flags in Dev. Is this all I need to do in the code to trigger this update (in addition to setting the version in the environment)?

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,13 +2,13 @@
 
 [[package]]
 name = "agate"
-version = "1.7.1"
+version = "1.9.1"
 description = "A data analysis library that is optimized for humans instead of machines."
 optional = false
 python-versions = "*"
 files = [
-    {file = "agate-1.7.1-py2.py3-none-any.whl", hash = "sha256:23f9f412f74f97b72f82b1525ab235cc816bc8c8525d968a091576a0dbc54a5f"},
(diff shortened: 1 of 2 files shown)


ian-r-rose:
You also need to update the `poetry.lock` as part of the process. Taking a closer look, I think I might have also made a mistake with the previous version specifier (I think we want `~1.8` rather than `~1.8.0`)

So a single command to update both the `pyproject.toml` and the `poetry.lock` would be:

```bash
poetry add "dbt-core@~1.8" "@dbt-snowflake@~1.8"
```

Of course, if we adopt `uv` instead of poetry, it would be a different (but similar!) command

summer-mothwood:
@ian-r-rose I ended up doing the following:
- manually changed the versions of the dbt-core and dbt-snowflake packages to ~1.8 in pyproject.toml
- ran `poetry lock` locally on this branch to update the poetry lock file
- staged and committed the poetry lock file changes

If that all looks right, I'll add these instructions to our docs as well. 

I couldn't get `poetry add "dbt-core@~1.8" "@dbt-snowflake@~1.8"` to run on a clean branch, as I kept getting an error on the dbt-snowflake package:
image

ian-r-rose:
Whoops, just a typo in that I added an extra `@` symbol in my comment above. Your changes look good to me!

ian-r-rose (approved):

Issue #352, opened 2024-08-06
Make slim CI work better


ian-r-rose:
We currently use "Slim CI" for this project, which (in theory), only runs models that are *affected* by a PR in CI. The intention here is that we can save time and money by just running the stuff that needs to be run. It works by comparing the code in the branch with that in the nightly production job, and identifying ones where there is a diff.

This is not working as intended because of the way we do some different logic depending upon whether we are "dev" or "prd". In particular, we select *less* data in dev for a lot of our larger models to improve the developer experience for those models. Unfortunately, this means that there _is_ a real difference between dev and prod, even if the pre-rendered model has not changed, thus creating some [false positives](https://docs.getdbt.com/reference/node-selection/state-comparison-caveats#false-positives) in our slim CI job.

I'm not sure of the best way to fix this right now.

jkarpen:
Assigning to @kengodleskidot to review and see if he has any thoughts on how to move forward on this one.

jkarpen:
We put this in the backlog until @ian-r-rose returns to get his input on how to handle this item.

jkarpen:
Summer will tackle this research project, will assign once she is added to the right GitHub team.

summer-mothwood:
Note: the dbt documentation page on state modified caveats have been updated a few times over the last few weeks, with some changes that might specifically address our issue -- but would require a version upgrade from dbt 1.7.18 (current Caltrans dbt version) to dbt 1.9+ or dbt Versionless. 


* https://github.com/dbt-labs/docs.getdbt.com/pull/6207

> Beginning in dbt Core 1.9, when you set the `state_modified_compare_vars` [behavior flag](/reference/global-configs/behavior-changes#behavior-change-flags) to `True` and a model uses a `var` or `env_var` in its definition, dbt will identify that lineage in such a way that it will include the model in `state:modified` when the `var` or `env_var` value has changed. 


* https://github.com/dbt-labs/docs.getdbt.com/commit/cb69cf640a0fa91b52488669fd28c3ecae860b1d

> You need to build the state directory using dbt v1.9 or higher, or [Versionless](/docs/dbt-versions/versionless-cloud) dbt Cloud, and you need to set `state_modified_compare_more_unrendered_values` to `true` within your dbt_project.yml.
[...]

PR #495, opened 2024-12-05
Summer versionless test

summer-mothwood:
Testing the CI job timing with the same change, now that the prod environment is working in Versionless and we've had a successful Nightly build. 

The hope here is that there is a noticeable difference in CI timing from the test I ran yesterday, and (more importantly, since timing might have been affected by the missing data from the detector source tables yesterday) that the logs show that CI is not rebuilding incremental models that had no changes made to them in this PR.

### DIFF
--- a/transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
+++ b/transform/models/intermediate/performance/int_performance__station_aadt_with_K_value.sql
@@ -480,5 +480,6 @@ select distinct
     aadt_8,
     k_30,
     k_50,
-    k_100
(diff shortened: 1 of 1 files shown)


summer-mothwood:
Hey! It worked! "Nothing to do" is the message I've been looking for. I'll update the [slim CI issue](https://github.com/cagov/caldata-mdsa-caltrans-pems/issues/352) w/ the before-and-after comparison once CI finishes running here. But this is great news:

image

ian-r-rose:
Woohoo!

PR #496, opened 2024-12-10
Added prevent_destroy flags to the two CH schemas

ram-kishore-odi:
Added prevent destroy flags for two important CalTrans schemas. Please review and let me know if you have questions / feedback.


ian-r-rose (approved):
I see a small pre-commit issue, but the content looks good to me!

PR #500, opened 2024-12-16
Updated is_trasient flag, pointed to the right commit for prd

ram-kishore-odi:
Can you please review the changes related to prd upgrade for the terraform snowflake provider for v 0.97 ?


ian-r-rose (approved):
Looks good, thanks @ram-kishore-odi!

PR #502, opened 2024-12-19
Update Models with gfactor Speed

### LINKED ISSUE #484: Update current speed in the clearinghouse using g-factor speed
As mentioned in today's sprint check-in, we need to update the speed in the clearinghouse for downstream models. @kengodleskidot Can you take a look at it once the updated g-factor speed is merged? It should be merged today.

kengodleskidot:
This PR makes the following changes:

- Update the detector_agg reference in the g-factor model from the int_clearinghouse__detector_agg_five_minutes_with_missing_rows to int_clearinghouse__detector_agg_five_minutes model
- In the int_clearinghouse__detector_agg_five_minutes_with_missing_rows model I added the g-factor speed data to detectors/timestamps where speed is not being provided by the field devices
- In the int_imputation__detector_agg_five_minutes and int_performance__detector_metrics_agg_five_minutes models remove the old speed calculation formula and notes
- YAML file updated describing the speed value being used in the int_clearinghouse__detector_agg_five_minutes_with_missing_rows model.

Resolves #484

### DIFF
--- a/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
+++ b/transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
@@ -41,6 +41,15 @@ spine as (
         on to_date(ts.timestamp_column) = dd.sample_date
 ),
 
+-- Add the model where gfactor speed has been calculated
+gfactor_speed as (
+    select
+        detector_id,
+        sample_timestamp,
+        imputed_speed
+    from {{ ref('int_clearinghouse__detector_g_factor_based_speed') }}
+),
+
 /* Join 5-minute aggregated data to the spine to get a table without missing rows */
 base as (
     select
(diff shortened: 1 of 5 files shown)


transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
```diff
@@ -88,6 +97,10 @@ base as (
                 to_date(spine.timestamp_column) < dmeta._valid_to
                 or dmeta._valid_to is null
             )
+    left join gfactor_speed as gs
+        on
+            agg.detector_id = gs.detector_id
+            and agg.sample_timestamp = gs.sample_timestamp
```
JamesSLogan, line 103: Note: adding another join to this already expensive model doesn't seems to hurt performance too much ([see query profile from CI here](https://app.snowflake.com/vsb79059/dse_caltrans_pems/#/compute/history/queries/01b924d7-0004-5d08-003e-3887014781ea/profile)), but let's keep our eyes on it.

kengodleskidot: The other alternatives we could investigate are creating a new model downstream of int_clearinghouse__detector_agg_five_minutes_with_missing_rows to add the g-factor speed or possibly add it to the int_imputation__detector_agg_five_minutes model.

thehanggit (approved):
Thank you @kengodleskidot. I think @JamesSLogan made a good point. Other than that, everything looks good. Let's have our own g-factor speed!

Issue #278, opened 2024-07-10
High Flow Values


kengodleskidot:
When reviewing aggregations of the 30-second (raw) data to 5-minutes we have come across observed flow values that are extremely high (500+ vehicles / lane / 5-minutes). Below is an analysis of the aggregated 5-minute data for detectors in lane 1 over a one-month time period as an example:
![image](https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/124397677/bd917715-53e0-4b5f-b6d3-0946d12a816c)

I then reviewed the data (Raw and 5-minute) in current PeMS to see how high flow values are being handled and observed the following:

- Nearly all the stations I reviewed with high flow values were flagged as Good based on the current diagnostic tests when in fact the data being provided appeared bad
- When comparing the raw data to the 5-minute data in the current PeMS, a variety of things were happening. In the example below the 5-minute data for lane 1 used observed flow data, lane 2 used normalized flow data, lanes 3 and 4 appeared to be using imputed data, all lanes for this station returned a Good status and the % observed value of 100% gives the impression that the values are all observed which they are not: 
Observed Data
![image](https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/124397677/faa06095-c1f4-4fa1-8502-c68a4356c2a8)
5-Minute Processed Data
![image](https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/124397677/95aeda06-392c-43c2-924d-5a2d80e97d0a)
Detector Health
![image](https://github.com/cagov/caldata-mdsa-caltrans-pems/assets/124397677/b2d4488f-31ed-47b8-8172-7b5a5d0c2eb5)

We do not currently have a diagnostic test for mainline and HOV stations to diagnose if high flow values at the detector level should result in a Bad detector status. To help identify issues associated with excessively high flow values I will create a new diagnostic test to check the number of high flow values for mainline and HOV lanes similar to what is being done for the High Occupancy diagnostic test. We will also want to incorporate logic in the final 5-minute aggregate table that looks at the observed flow value and replaces excessively high flow values with either the imputed value or the max capacity value at the lane level.

kengodleskidot:
@mmmiah and I will review the 'int_performance__detector_metrics_agg_five_minutes' model to see how we incorporate checking for high flow values and how to handle this scenario with imputed data.
[...]

PR #506, opened 2025-01-02
High Flow and Occupancy Detection and Correction

thehanggit:
This is related to Issue #278


ian-r-rose (commented):
From a high-level, this logic seems reasonable to me. I do wonder if it would be worth breaking out the outlier removal into a separate intermediate model: I could imagine that more than one "performance" model would want to use the version with outliers removed. What do you think @thehanggit ?

thehanggit:
> From a high-level, this logic seems reasonable to me. I do wonder if it would be worth breaking out the outlier removal into a separate intermediate model: I could imagine that more than one "performance" model would want to use the version with outliers removed. What do you think @thehanggit ?

Thank you @ian-r-rose for taking a look! I talked with @kengodleskidot just now and we may want to detect and fix the true outliers in the clearinghouse. As the outliers in the performance model might be attributed to imputation, which are the "fake" outliers, and we want to leave that part to the imputation model. I will develop a separate outlier removal in the clearinghouse folder because it would also influence the g-factor speed calculation.

transform/models/intermediate/performance/int_performance__detector_metrics_agg_five_minutes.sql
```diff
@@ -35,6 +35,64 @@ five_minute_agg as (
...
+        detector_id,
+        date_trunc('month', sample_date) as month,
+        avg(volume_sum) as volume_mean,
+        stddev(volume_sum) as volume_stddev,
+        -- consider using max_capacity
+        percentile_cont(0.95) within group (order by volume_sum) as volume_95th,
+        percentile_cont(0.95) within group (order by occupancy_avg) as occupancy_95th
+    from five_minute_agg
+    group by detector_id, date_trunc('month', sample_date)
+),
```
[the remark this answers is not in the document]
thehanggit: In that case, I can use the preceding month's statistics (mean, std) to predict outlier for the current month.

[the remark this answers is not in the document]
ian-r-rose: I think that's fine, but the incremental logic would have to change: the logic here doesn't really work for any date because you are not actually ever computing monthly stats (`five_minute_agg` only has two days of data in it)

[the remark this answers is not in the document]
kengodleskidot: The updated approach will be to create a separate high flow outlier detection model in the clearinghouse schema that will need to account for the incremental logic scenario @ian-r-rose mentions in the previous comment. I can see a scenario where incrementality is not needed but we will see what @thehanggit comes up with. Looking forward to seeing the update!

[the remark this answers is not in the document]
ian-r-rose: Sounds good!

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_outlier_agg_five_minutes.sql
```diff
@@ -0,0 +1,109 @@
...
+with
+five_minute_agg_lastmonth as (
+    select
+        detector_id,
+        sample_date,
+        volume_sum,
+        occupancy_avg
+    from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
+    where
+        sample_date >= dateadd(month, -1, date_trunc('month', current_date))
```
kengodleskidot, line 20: I believe one of these values is a timestamp and the other is a date. Do you run into any issues when preforming this comparison?

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_outlier_agg_five_minutes.sql
```diff
@@ -0,0 +1,109 @@
...
+
+-- impute detected outliers
+outlier_removed_data as (
+    select
+        fa.*,
+        -- update volume_sum if it's an outlier
+        case
+            when
+                fa.volume_sum - ms.volume_mean / nullifzero(ms.volume_stddev) > 3
+                then coalesce(ms.volume_95th, 173)
```
kengodleskidot, line 77: I recommend adding a note here explaining why a value of 173 is being used.

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_outlier_agg_five_minutes.sql
```diff
@@ -0,0 +1,109 @@
...
+            when
+                fa.volume_sum - ms.volume_mean / nullifzero(ms.volume_stddev) > 3
+                then 'outlier'
+            else 'observed data'
+        end as volume_label,
+        -- update occupancy if it's an outlier
+        case
+            when
+                fa.occupancy_avg > ms.occupancy_95th
+                then coalesce(ms.occupancy_95th, 0.8)
```
kengodleskidot, line 91: Adding a similar note here on why 0.8 is being used would be beneficial.

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_outlier_agg_five_minutes.sql
```diff
@@ -0,0 +1,109 @@
...
+        and station_type in ('ML', 'HV')
+),
+
+-- get all good detectors
+good_detectors as (
+    select
+        detector_id,
+        sample_date
+    from {{ ref('int_diagnostics__detector_status') }}
+    where status = 'Good'
```
kengodleskidot, line 31: Should you add a date timeframe to this where statement similar to the previous CTE? Otherwise, you are probably grabbing a very large data set that may affect your model performance.

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_g_factor_based_speed.sql
```diff
@@ -16,11 +16,10 @@ detector_agg as (
         station_id,
         lane,
         station_type,
-        volume_sum,
-        occupancy_avg,
-        speed_weighted,
```
kengodleskidot, line 21: I see you are removing speed_weighted from this model, will this cause any downstream issues?

kengodleskidot (approved):
Nice work @thehanggit! I don't want my comments to hold up this PR since I do not believe it would change the output of the model but feel free to reach out if you would like to discuss any of my comments.

ian-r-rose:
@thehanggit this model timed out after taking 10 hours today -- I think we may need disable it and take a closer look at the performance characteristics.

thehanggit:
@ian-r-rose I checked the query profile by running a test in worksheet. It turns out the aggregate (not sure but it may relate to the var calculate?) is expensive. I switched from monthly data to weekly data, reducing the size by half. Another way is to use 95th percentile only to detect outliers, which I tested before and they also have a good performance for high flow value detection.

image
image

PR #512, opened 2025-01-08
disable poetry package mode

JamesSLogan:
It looks like poetry 2.0 is out and our pre-commit workflow (at a minimum) is broken. I think the clearest solution is to call out, explicitly, that we're using poetry for dependency management only.


ian-r-rose (approved):
Thanks @JamesSLogan

PR #513, opened 2025-01-09
yaml file edits for data grain, adding data tests

summer-mothwood:
Added data tests, made data grain explicit for certain models, made other copy edits and small fixes to yaml files across the entire pipeline

### DIFF
--- a/transform/models/marts/imputation/_imputation.yml
+++ b/transform/models/marts/imputation/_imputation.yml
@@ -4,8 +4,17 @@ models:
       This model summarize the pentration of imputation by each method.
       It provides the total number of samples for five minutes in station
       level as well as percetage of observed and imputed by each method
-      by sample date.
+      by sample date. This model is unique at the level of SAMPLE_DATE + DETECTOR_ID.
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - SAMPLE_DATE
+            - DETECTOR_ID
     columns:
+      - name: DETECTOR_ID
+        description: An integer value that uniquely indentifies a station.
+        data_tests:
+          - not_null
       - name: STATION_ID
         description: |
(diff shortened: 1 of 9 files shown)


transform/models/marts/imputation/_imputation.yml
```diff
@@ -66,17 +77,30 @@ models:
...
-      and 'global' respectively. The model reported the observed values for all good detetcors.
+      This model filters the results of the imputation_detector_imputed_agg_five_minutes model from the transform
+      step to detectors with station_type in of either 'ML' or 'HV' from the past 4 days and adds the county name
+      to the results. This model is unique at the level of SAMPLE_TIMESTAMP + DETECTOR_ID.
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - SAMPLE_TIMESTAMP
+            - DETECTOR_ID
+            - COUNTY
```
ian-r-rose, line 88: Do we need to include `COUNTY` here? I would think that timestamp and detector ID should be plenty for uniqueness

ian-r-rose (commented):
I'm worried that the five-minute aggregation models are too large to perform uniqueness tests (uniqueness tests are really expensive on big data, which is why Snowflake doesn't actually respect uniqueness constraints). Each one is taking ~20 minutes to run, and in CI we are only selecting seven days of data, so I suspect in prod they would take a very long time.

CI timing [here](https://california.getdbt.com/deploy/6/projects/10/runs/7630/?tabId=model-timing)

ian-r-rose:
Overall this is looking great, and I'm excited to have this in!

transform/models/marts/imputation/_imputation.yml
summer-mothwood: Yes, good call!

PR #514, opened 2025-01-09
Add source tests

### LINKED ISSUE #508: Test that config tables are not empty
Problems in the config table pipeline can result in tables that are still there, but empty. We should write a generic test that checks whether a table is empty and use it on the following tables:

* `CONTROLLER_CONFIG`
* `CONTROLLER_CONFIG_LOG`
* `DETECTOR_CONFIG`
* `DETECTOR_CONFIG_LOG`
* `STATION_CONFIG`
* `STATION_CONFIG_LOG`

### LINKED ISSUE #509: Test that config tables don't have duplicates
Problems in the data relay pipelines for the config tables have resulted in the same data being uploaded multiple times per day. We should ensure that the following tables don't have duplicate records:

* `CONTROLLER_CONFIG`
* `CONTROLLER_CONFIG_LOG`
* `DETECTOR_CONFIG`
* `DETECTOR_CONFIG_LOG`
* `STATION_CONFIG`
* `STATION_CONFIG_LOG`

The `_LOG` tables *do* have multiple entries per device, but *should* be unique when you look at the combination of device ID and date. In that case, I think we should be able to use the generic [`unique_combination_of_columns`](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#unique_combination_of_columns-source) test.

JamesSLogan:
Fixes #508
Fixes #509

### DIFF
--- a/transform/models/_sources.yml
+++ b/transform/models/_sources.yml
@@ -291,6 +291,8 @@ sources:
         freshness: null
         tests:
           - not_empty
+          - unique:
+              column_name: controller_id
         columns:
           - name: meta
             description: Metadata from the data relay scripts.
@@ -314,6 +316,10 @@ sources:
           with the `controller_config` table to get a full history of controller metadata.
         tests:
           - not_empty
+          - dbt_utils.unique_combination_of_columns:
+              combination_of_columns:
+                - controller_id
+                - time_id
         columns:
           - name: meta
             description: Metadata from the data relay scripts.
@@ -348,6 +354,8 @@ sources:
         freshness: null
         tests:
           - not_empty
+          - unique:
+              column_name: station_id
         columns:
           - name: meta
             description: Metadata from the data relay scripts.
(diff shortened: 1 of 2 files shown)


transform/tests/generic/not_empty.sql
ian-r-rose, line 1: I was surprised the other day to see that `dbt_utils` didn't already have a test for this!

ian-r-rose (approved):
Looks good, thanks @JamesSLogan

transform/models/_sources.yml
```diff
@@ -289,6 +289,10 @@ sources:
           controller might have a station for the mainline as well as a station
           for an onramp.
         freshness: null
+        tests:
```
ian-r-rose, line 292: If I understand correctly from @summer-mothwood, this key is deprecated, and we should be referring to these as `data_tests` now

ian-r-rose:
Sorry @JamesSLogan can your merge or rebase? Looks like we've picked up a conflict with my PR

JamesSLogan:
Done - the merge wasn't particularly complex, but could you double check?

ian-r-rose (approved):
Thanks!

PR #516, opened 2025-01-13
Add Copy History tests to db96 Sources

summer-mothwood:
Created a generic test to check if last load time in Snowflake Copy History for a given model is within the last 24 hours.

Applied that test to these source tables:

```
    CONTROLLER_CONFIG
    CONTROLLER_CONFIG_LOG
    DETECTOR_CONFIG
    DETECTOR_CONFIG_LOG
    STATION_CONFIG
    STATION_CONFIG_LOG
```

### DIFF
--- a/transform/tests/generic/copy_history_24h.sql
+++ b/transform/tests/generic/copy_history_24h.sql
@@ -1,0 +1,25 @@
+{% test copy_history_24h(model) %}
+
+/* This grabs the model's available copy history (past 14 days) from Snowflake
+see: https://docs.snowflake.com/en/sql-reference/functions/copy_history
+*/
+with validation as (
+   select LAST_LOAD_TIME
+   from table(
+    INFORMATION_SCHEMA.COPY_HISTORY(
+        table_name => '{{ model }}',
+        start_time=> DATEADD(days, -14, CURRENT_TIMESTAMP())
+        )
+    )
+),
+
+-- This checks if the last recorded load time was more than 24 hours ago
+validation_errors as (
+   select LAST_LOAD_TIME
+   from validation
+   where LAST_LOAD_TIME < DATEADD(days, -1, CURRENT_TIMESTAMP())
+)
+
+select * from validation_errors
+
+{% endtest %}
--- a/transform/models/_sources.yml
+++ b/transform/models/_sources.yml
@@ -282,6 +282,7 @@ sources:
           for an onramp.
         data_tests:
           - not_empty
+          - copy_history_24h
           - unique:
               column_name: controller_id
         columns:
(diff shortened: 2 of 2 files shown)


transform/tests/generic/copy_history_24h.sql
```diff
@@ -0,0 +1,25 @@
+{% test copy_history_24h(model) %}
+
+/* This grabs the model's available copy history (past 14 days) from Snowflake
+see: https://docs.snowflake.com/en/sql-reference/functions/copy_history
+*/
+with validation as (
+   select LAST_LOAD_TIME
+   from table(
+    INFORMATION_SCHEMA.COPY_HISTORY(
```
ian-r-rose, line 9: Does `INFORMATION_SCHEMA` need to be qualified with a database name? i.e., something like `{{ model.database }}.information_schema.copy_history`?

ian-r-rose (commented):
Nice! Just one question, but code overall looks good to me

transform/tests/generic/copy_history_24h.sql
summer-mothwood: Good question, as far as I can tell it does not, but tbh I still don't think I understand all the intricacies of how dbt + snowflake talk to each other in CI/Dev vs Prod. So I'm not sure how else to test, but I did do the following:

I ran this on detector_config exactly as is except reversing the date test (looking for last load time >= today - 1 day), and it successfully failed, so I know this syntax works to correctly pull data from a given model's copy history. 

I also referenced this utils package someone wrote for inspiration https://github.com/entechlog/dbt-snow-utils/blob/main/models/staging/snowpipe/stg_snowpipe__copy_history.sql , and while they are setting the database name in order to run a check that they're in a valid database, it does not figure into the copy history selection query. 

lmk what you think!

ian-r-rose: I'm a little surprised it didn't seem to matter. Our CI user uses `transform_dev` as the default database, but these tables are in `raw_prd`. What happens when we query for the copy history of `raw_prd.db96.controller_config` using the `transform_dev.information_schema.copy_history` function? Seemingly, it works, but it seems odd to me!

ian-r-rose: Sure enough, this works in a worksheet:

```sql
select *
from table(
    transform_dev.information_schema.copy_history(
        TABLE_NAME=>'RAW_PRD.DB96.CONTROLLER_CONFIG_LOG',
        START_TIME=> DATEADD(days, -14, CURRENT_TIMESTAMP())
    )
);
```
Weird!

ian-r-rose (approved):
Looks good!

Issue #521, opened 2025-01-16
Unpivot columns by thresholds


thehanggit:
In the mart folder, there are similar metrics generated by different values of speed thresholds, which resulted into different columns. We may consider unpivoting these similar columns by adding a threshold column and integrating them into one column for each type of metrics (delay, extent, etc).

This needs to be investigated by Ian before action for the reason that it would duplicate rows significantly.

PR #522, opened 2025-01-21
Unpivot columns by thresholds

thehanggit:
This is related to #521

### DIFF
--- a/transform/models/marts/performance/performance__agg_yearly_city.sql
+++ b/transform/models/marts/performance/performance__agg_yearly_city.sql
@@ -8,7 +8,7 @@ with station_daily_data as (
     from {{ ref('int_performance__station_metrics_agg_daily') }}
 ),
 
--- now aggregate daily volume, occupancy and speed to weekly
+-- now aggregate daily volume, occupancy and speed to yearly
 spatial_metrics as (
     select
         city,
@@ -20,23 +20,7 @@ spatial_metrics as (
         sum(daily_vht) as yearly_vht,
         yearly_vmt / nullifzero(yearly_vht) as yearly_q_value,
         -- travel time
-        60 / nullifzero(yearly_q_value) as yearly_tti,
-        {% for value in var("V_t") %}
-            sum(delay_{{ value }}_mph)
-                as delay_{{ value }}_mph
-            {% if not loop.last %}
-                ,
-            {% endif %}
-
-        {% endfor %},
-        {% for value in var("V_t") %}
-            sum(lost_productivity_{{ value }}_mph)
-                as lost_productivity_{{ value }}_mph
-            {% if not loop.last %}
-                ,
-            {% endif %}
-
-        {% endfor %}
+        60 / nullifzero(yearly_q_value) as yearly_tti
     from station_daily_data
     group by
         city, sample_year
--- a/transform/models/marts/performance/performance__agg_yearly_county.sql
+++ b/transform/models/marts/performance/performance__agg_yearly_county.sql
@@ -20,23 +20,7 @@ spatial_metrics as (
         sum(daily_vht) as yearly_vht,
         yearly_vmt / nullifzero(yearly_vht) as yearly_q_value,
         -- travel time
-        60 / nullifzero(yearly_q_value) as yearly_tti,
-        {% for value in var("V_t") %}
-            sum(delay_{{ value }}_mph)
-                as delay_{{ value }}_mph
-            {% if not loop.last %}
-                ,
-            {% endif %}
-
-        {% endfor %},
-        {% for value in var("V_t") %}
-            sum(lost_productivity_{{ value }}_mph)
-                as lost_productivity_{{ value }}_mph
-            {% if not loop.last %}
-                ,
-            {% endif %}
-
-        {% endfor %}
+        60 / nullifzero(yearly_q_value) as yearly_tti
     from station_daily_data
     group by
         county, sample_year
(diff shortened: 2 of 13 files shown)


transform/models/marts/performance/performance__unpivot_agg_yearly_district.sql
```diff
@@ -0,0 +1,101 @@
...
+            )
+        )
+    )
+),
+
+unpivot_lost_productivity as (
+    select
+        district,
+        sample_year,
+        regexp_substr(metric, '([0-9])+', 1, 1) as target_speed,
```
ian-r-rose, line 64: I don't *love* using a regex to get the speed here, but I'm not sure if I can think of a better way right now. At the very least, I think it makes sense to cast this as a numeric value, as it may be useful in downstream analyses to actually do math on the target speeds.

transform/models/marts/performance/performance__unpivot_agg_yearly_district.sql
```diff
@@ -0,0 +1,101 @@
...
+        d.delay,
+        lp.lost_productivity
+    from
+        unpivot_delay as d
+    left join
+        unpivot_lost_productivity as lp
+        on
+            d.district = lp.district
+            and d.sample_year = lp.sample_year
+            and d.target_speed = lp.target_speed
```
ian-r-rose, line 96: In #124 I worked *really* hard to try to get the `unpivot` function to allow unpivoting multiple values so that this join wasn't necessary. Because we shouldn't need to do this (expensive) join! These values both come from the same `spatial_metrics` table, and you're not doing any reordering, so we shouldn't need to do any joining at all.

At the end of the day, I never really figured it out, and ultimately this is what led me to the `union all` approach instead of using `unpivot`.

What you are doing here makes sense, and looks correct to me, but I suggest keeping an eye on performance, and if things start to look bad, we might want to consider the `union all` approach used for unpivoting the lane data.

transform/models/marts/performance/performance__unpivot_agg_yearly_freeway_type_direction.sql
ian-r-rose, line 1: This file is one where it might make sense to restructure things so that you don't have to unpivot. Instead you could compute a separate `spatial_metrics` CTE for each delay and then union them. As it is, you are first constructing the CTE so it is in wide-format, and then converting it to long.

ian-r-rose (commented):
@thehanggit this is looking really nice. I have some suggestions for things to keep an eye out for with performance, but the overall code looks correct to me. I know you also had some things you wanted to do based on our conversation with @ZhenyuZhu-Caltrans, so let me know when this is ready for another look!

thehanggit:
> @thehanggit this is looking really nice. I have some suggestions for things to keep an eye out for with performance, but the overall code looks correct to me. I know you also had some things you wanted to do based on our conversation with @ZhenyuZhu-Caltrans, so let me know when this is ready for another look!

Really appreciated your comments @ian-r-rose!  I will take some time to digest and modify the codes based on your comments and invite you for another round of review!

transform/models/marts/performance/performance__unpivot_agg_yearly_district.sql
thehanggit: @ian-r-rose I have revised the code by using union all to approach and demonstrate it in `performance__unpivot_agg_yearly_district.sql`. Please check it and if it makes sense to you, I will change other tables accordingly.

transform/models/marts/performance/performance__unpivot_agg_yearly_district.sql
thehanggit: Have avoided using regex function and demonstrated it in `performance__unpivot_agg_yearly_district.sql`

transform/models/marts/performance/performance__station_metrics_agg_daily.sql
mmmiah, line 1: do we want city name instead of city code for this model and also County Short name?

thehanggit: For sure, I will address them in another issue. This one is mainly for unpivot.

transform/models/marts/performance/performance__bottleneck_agg_daily.sql
mmmiah, line 1: add county Short name aslo

transform/models/marts/performance/performance__agg_yearly_county.sql
mmmiah, line 1: Consider adding county short name also

transform/models/marts/performance/performance__agg_yearly_city.sql
mmmiah, line 1: Consider replacing city code with city name

transform/models/marts/performance/performance__station_metrics_agg_monthly.sql
mmmiah, line 1: replace city code with city name and also add county_abb column

transform/models/marts/performance/performance__station_metrics_unpivot_agg_monthly.sql
mmmiah, line 1: Consider replacing city code with name and adding county short name

transform/models/marts/performance/performance__unpivot_agg_yearly_city.sql
mmmiah, line 1: Consider replacing city code with name and adding county short name

transform/models/marts/performance/performance__unpivot_agg_yearly_county.sql
mmmiah, line 1: add county short name

mmmiah (commented):
Hang, looks good to me, I have added some minor comments, please address those. We are very close to approve this PR. Thanks for doing all these.

thehanggit:
> Hang, looks good to me, I have added some minor comments, please address those. We are very close to approve this PR. Thanks for doing all these.

@mmmiah As I mentioned, I will address them in #525. Please approve it if everything else looks good. Thank you

mmmiah (approved):

Issue #526, opened 2025-01-27
Create more aggregation tables based on different spatial and temporal levels


thehanggit:
- [ ] Yanlin will conduct yearly station aggregation
- [x] Hang will conduct weekly and yearly bottleneck aggregation
- [x]  Hang will conduct county, district, state, and city level aggregation for daily, weekly, monthly station data

PR #530, opened 2025-01-28
created daily, weekly, and monthly table in mart for bottlenecks

thehanggit:
This is related to #526 .


YanlinQi-BIA (approved):
Everything looks pretty good! @thehanggit

PR #532, opened 2025-01-30
Name Check for `marts-performance`

YanlinQi-BIA:
Reviewed and updated the `marts-performance` `.sql` files according to Ian's rule #241 : 
`{model stage}_{category}__{device type}_agg_{temporal aggregation}_{spatial_aggregation}`


thehanggit:
Thank you @YanlinQi-BIA for your efforts! There are minor things you may need pay attention, please take a look. I would suggest not having "station" info if there is a spatial aggregation to avoid confusion. But I agree with removing metrics to shorten the length of table names.

YanlinQi-BIA:
> Thank you @YanlinQi-BIA for your efforts! There are minor things you may need pay attention, please take a look. I would suggest not having "station" info if there is a spatial aggregation to avoid confusion. But I agree with removing metrics to shorten the length of table names.

Hi @thehanggit, thank you for reviewing. For `station` info, I referenced #241 and followed this rule:
`device type: For VDS data this would usually be "controller", "station", or "detector", and indicates the "grain" of the model. For instance, most of our VDS tables are at the "detector" level because they include a lane column. But when the lanes are aggregated up to the station level, it would be "station". Note that this could also be considered a spatial aggregation, but in most cases I felt it was important enough information to have a dedicated part in the name.`

I reviewed the code for each model and included "station" in the name when the model's grain matched. Let me know if we decide to exclude this part.

thehanggit:
@YanlinQi-BIA Looks great! Please feel free to merge it.

thehanggit (approved):

Issue #511, opened 2025-01-08
Constant Occupancy Detector Status Evaluation


kengodleskidot:
To confirm our constant occupancy model [int_diagnostics__constant_occupancy.sql](https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/main/transform/models/intermediate/diagnostics/int_diagnostics__constant_occupancy.sql) is correctly diagnosing detectors we can spot check the Tableau dashboard (https://tableau.dot.ca.gov/t/Construction/views/PeMS_Detector_Status_QC-Detailed/PeMSDetectorDetailsQualityCheck) and see which detectors on which dates show a Constant status in the existing system while the modern system has a different status (typically "Good")

thehanggit:
@kengodleskidot I dived into the diagnostic code to check the difference for "good" and "constant". I noticed that we used `min` function at line 58 in [int_diagnostics__constant_occupancy.sql](https://github.com/cagov/caldata-mdsa-caltrans-pems/blob/main/transform/models/intermediate/diagnostics/int_diagnostics__constant_occupancy.sql) for `occupancy_delta`, which means as long as there is a 4-hour window remaining constant within that day, it would be diagnosed as `constant`. For example, the `detector_id` = 440253 in modern PeMS is diagnosed as `Constant`, whereas the current PeMS is diagnosed as `Good` on 2/13/2023.

![Image](https://github.com/user-attachments/assets/2acf0a67-ab0f-4eea-b222-c3a6d5308158)

However, when I pulled out the `sum_occupancy_delta` table, the values `abs_cal_occupancy_delta_summed` column vary starting from 12:25:00, which means the occupancy has variance at that time. Then if we use `min` function, then **min(abs_cal_occupancy_delta_summed) = 0** and we treat the detector status constant. That is one possible issue we can look into.

Another issue might be: The rolling window function calculates a time window every 5 minute, should it be the case that for a time window 5:00 - 21:00, we calculate 4 data points from 5:00-9:00, 9:00-13:00, 13:00-17:00, 17:00-21:00 as metrics for detector status?

kengodleskidot:
[...]

PR #535, opened 2025-01-30
Fix logic and threshold for diagnose

thehanggit:
This is related to #511


kengodleskidot (approved):
Looks great, thank you @thehanggit!

Issue #529, opened 2025-01-28
Investigate Freshness Job Notifications


summer-mothwood:
I've been getting Warning emails on the Freshness job for a couple days, but James and Ian haven't. This ticket is to review the notification settings on that job to ensure the rest of the team is getting these warnings.

JamesSLogan:
Presently, dbt gives us this deprecation-related warning on each run of the `source freshness` job:
> [WARNING]: In a future version of dbt, the `source freshness` command
will start running `on-run-start` and `on-run-end` hooks by default. For more
information: https://docs.getdbt.com/reference/global-configs/legacy-behaviors

This warning does not go away even though we have the `source_freshness_run_project_hooks` flag set to False. Disabling our start hooks altogether does eliminate the warning, but I can't find another way to suppress it. [Relevant source code](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/freshness.py#L275). 

So, @summer-mothwood I think you may want to adjust your notification settings since I don't see the warning going away any time soon.

jkarpen:
Closing this as there is no action to take at this time.

PR #539, opened 2025-02-06
fix a few warnings

JamesSLogan:
This PR makes a few miscellaneous fixes noticed while looking at #529 (which is otherwise unrelated to this PR). 

I also noticed some warnings about models in `models/marts/performance/_performance.yml` but did not address them here. I'll open a follow-on issue to fix those since those models are in flux currently (#538).


summer-mothwood (approved):
thanks James!

PR #545, opened 2025-02-20
Iceberg materialization test

### LINKED ISSUE #503: Investigate using external Iceberg tables to publish larger marts
Our current approach to publishing marts to S3 works reasonably well for smaller tables (less than a few hundred MB), but does not work as well when they are in the hundreds or thousands of GB. This includes most of the 5-minute aggregation tables.

One possibility for publishing those tables is to use the new(ish) [external iceberg](https://docs.snowflake.com/en/user-guide/tables-iceberg) format. This has a few advantages:

1. Publish to an open S3 bucket that we own
2. Open source format (bring your own execution engine)
3. Based on parquet
4. Can be based on the same source of truth as what's in the data warehouse. Snowflake and external compute engines would be querying exactly the same data.

ian-r-rose:
Fixes #503 

A few implementation notes:

1. Rather than unconditionally unloading everything in the `marts` folder as partitioned parquet files, I apply the logic on the per-subdirectory level. Most subdirectories intended to have small, summary-level marts, but I've created a new `iceberg` subdirectory which is configured to unload the tables in iceberg format.
2. The iceberg format supports ms-level timestamps, which means we have to cast timestamps to that format using, e.g. `::timestamp_ntz(6)`
3. Incrementality for the iceberg table is supported in theory, but it is currently blocked by [this](https://github.com/dbt-labs/dbt-snowflake/pull/1293) bug in dbt's Snowflake adapter. I've tested the fix locally, and it works, but we wouldn't be able to use it in dbt Cloud until it's merged and published.


ian-r-rose:
Closing this PR for now, but I think we should revisit if the above upstream bug is fixed

PR #548, opened 2025-02-20
Fix unpivoting issues

thehanggit:
Found out the monthly bottleneck table is unpivot. I modified the code based on the daily and weekly bottleneck sql.


mmmiah (approved):
looks good. Thanks.

Issue #550, opened 2025-02-26
Recall partitioned tables in S3


thehanggit:
Data ingestion from S3 to our Streamlit web page can handle datasets of at least 20MB without noticeable latency when accessed via URL. So we set up 20MB as a threshold for users each time they create a dashboard based on data in marts. 

The issue here is: some tables, which are partitioned automatically by year and month like `performance_station_metrics_agg_daily` include suffix as show.

![Image](https://github.com/user-attachments/assets/cc8599bc-e7a8-49fe-b4c0-105148887864)

To build automatic pipelines that use URLs to create dashboards, we need a reliable method to extract this suffix. Could you advise on how we can obtain this suffix programmatically so that our pipelines can automatically reference the correct parquet file?

thehanggit:
Since the prefix is unique, we are able to locate these files and start to build the dashboard.

thehanggit:
Controlling each partition file within 20 MB considering 20 year's data is important for dashboard visualization. I will conduct a thorough analysis on each table to have a right partition mechanism and get all the tables ready for Streamlit data ingestion.

PR #551, opened 2025-03-03
clean data table by appropriate partition

thehanggit:
This is related to #550.

The goal here is to partition every table in marts/performance to ensure that the sizes of them are limited to 20-30 MB. Typically, we use temporal segmentation: 'year', 'month', etc.


thehanggit:
> when you are removing the partition, does it mean that this table is already less than 30MB and will not be bigger in size in future. I can see that you added the partition for couple of CTEs while mostly you removed the partition for others?

Correct! Most the city, county, and district based aggregation tables are small enough. We don't need to partition them again. I checked every table to make sure them in a reasonable size and have them partitioned if necessary.

mmmiah (approved):
Sounds good. Thanks for clarification.

thehanggit:
> Sounds good. Thanks for clarification.

Thank  you! @mmmiah

PR #552, opened 2025-03-19
Iceberg materialization test

ian-r-rose:
@thehanggit I'm reopening this with some additional experimentation around iceberg metadata. The main issue is that there can be many different versions of the iceberg metadata (in general, each change to the table results in a new version of the metadata). Snowflake keeps track of which version is the most recent, but it's not always easy to determine the correct one from a different tool trying to query the iceberg table.

This creates a new stored procedure that unloads a JSON blob to the `iceberg` folder in S3 that lists the correct metadata file for each iceberg table in the marts database. I was then able to use that directly in the following sample duckdb script:

```sql
-- Load the extensions allowing for S3 access and Iceberg
load iceberg;
load httpfs;

set variable name = 'IMPUTATION__DETECTOR_IMPUTED_AGG_FIVE_MINUTES';
set variable schema = 'DBT_IROSE_IMPUTATION';

-- Since this script doesn't manage the iceberg tables, it doesn't know
-- which version of the metadata is the correct one. We could guess based
-- on heuristics for file names or modification dates, but that is extremely
-- unreliable. So we unload a special file into the iceberg directory that
-- tracks the most recent version of the metadata for each iceberg table
-- (as of the time the current_table_versions.json file was written)
-- This grabs the metadata path for the table we care about.
set variable meta = (
    select metadata from read_json(
        "s3://caltrans-pems-dev-us-west-2-marts/iceberg/current_table_versions.json"
    )
    where name = getvariable('name') and schema = getvariable('schema')
);


-- Actually make a query with no authentication!
select
    station_id,
    sample_date,
    sum(volume_sum) as volume,
    count_if(
        volume_imputation_method in ('local', 'regional', 'global', 'local_avg', 'regional_avg')
    )/count(*) as pct_imputed,
from iceberg_scan(getvariable('meta'))
where sample_date > '2025-03-10' and sample_date < '2025-03-17'
group by station_id, sample_date
order by station_id, sample_date;
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -26,100 +26,105 @@ test = ["PyICU (>=2.4.2)", "backports.zoneinfo", "coverage (>=3.7.1)", "cssselec
 
 [[package]]
 name = "aiohappyeyeballs"
-version = "2.4.3"
+version = "2.6.1"
 description = "Happy Eyeballs for asyncio"
 optional = false
-python-versions = ">=3.8"
+python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "aiohappyeyeballs-2.4.3-py3-none-any.whl", hash = "sha256:8a7a83727b2756f394ab2895ea0765a0a8c475e3c71e98d43d76f22b4b435572"},
-    {file = "aiohappyeyeballs-2.4.3.tar.gz", hash = "sha256:75cf88a15106a5002a8eb1dab212525c00d1f4c0fa96e551c9fbe6f09a621586"},
+    {file = "aiohappyeyeballs-2.6.1-py3-none-any.whl", hash = "sha256:f349ba8f4b75cb25c99c5c2d84e997e485204d2902a9597802b0371f09331fb8"},
+    {file = "aiohappyeyeballs-2.6.1.tar.gz", hash = "sha256:c3f9d0113123803ccadfdf3f0faa505bc78e6a72d1cc4806cbd719826e943558"},
 ]
 
 [[package]]
 name = "aiohttp"
-version = "3.11.7"
+version = "3.11.18"
 description = "Async http client/server framework (asyncio)"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "aiohttp-3.11.7-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:8bedb1f6cb919af3b6353921c71281b1491f948ca64408871465d889b4ee1b66"},
-    {file = "aiohttp-3.11.7-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:f5022504adab881e2d801a88b748ea63f2a9d130e0b2c430824682a96f6534be"},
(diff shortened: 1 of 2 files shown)


thehanggit:
> @thehanggit I'm reopening this with some additional experimentation around iceberg metadata. The main issue is that there can be many different versions of the iceberg metadata (in general, each change to the table results in a new version of the metadata). Snowflake keeps track of which version is the most recent, but it's not always easy to determine the correct one from a different tool trying to query the iceberg table.
> 
> This creates a new stored procedure that unloads a JSON blob to the `iceberg` folder in S3 that lists the correct metadata file for each iceberg table in the marts database. I was then able to use that directly in the following sample duckdb script:
> 
> ```sql
> -- Load the extensions allowing for S3 access and Iceberg
> load iceberg;
> load httpfs;
> 
> set variable name = 'IMPUTATION__DETECTOR_IMPUTED_AGG_FIVE_MINUTES';
> set variable schema = 'DBT_IROSE_IMPUTATION';
> 
> -- Since this script doesn't manage the iceberg tables, it doesn't know
> -- which version of the metadata is the correct one. We could guess based
> -- on heuristics for file names or modification dates, but that is extremely
> -- unreliable. So we unload a special file into the iceberg directory that
> -- tracks the most recent version of the metadata for each iceberg table
> -- (as of the time the current_table_versions.json file was written)
> -- This grabs the metadata path for the table we care about.
> set variable meta = (
>     select metadata from read_json(
>         "s3://caltrans-pems-dev-us-west-2-marts/iceberg/current_table_versions.json"
>     )
>     where name = getvariable('name') and schema = getvariable('schema')
> );
> 
> 
> -- Actually make a query with no authentication!
> select
>     station_id,
>     sample_date,
>     sum(volume_sum) as volume,
>     count_if(
>         volume_imputation_method in ('local', 'regional', 'global', 'local_avg', 'regional_avg')
>     )/count(*) as pct_imputed,
> from iceberg_scan(getvariable('meta'))
> where sample_date > '2025-03-10' and sample_date < '2025-03-17'
> group by station_id, sample_date
> order by station_id, sample_date;
> ```

@ian-r-rose  Thank you Ian! One question here is: does this json file stores all iceberg tables' metadata information? And we can use such one json file to find all most updated metadata information by choosing variables for the table name and schema right.

ian-r-rose:
> One question here is: does this json file stores all iceberg tables' metadata information? And we can use such one json file to find all most updated metadata information by choosing variables for the table name and schema right.

Yes, exactly!


I'm attaching an example file that was created as part of this PR (which is also hosted in S3, as you can see in the above script):

[current_table_versions.json](https://github.com/user-attachments/files/19352246/current_table_versions.json)

thehanggit:
> > One question here is: does this json file stores all iceberg tables' metadata information? And we can use such one json file to find all most updated metadata information by choosing variables for the table name and schema right.
> 
> Yes, exactly!
> 
> I'm attaching an example file that was created as part of this PR (which is also hosted in S3, as you can see in the above script):
> 
> [current_table_versions.json](https://github.com/user-attachments/files/19352246/current_table_versions.json)

Thanks for clarification! I will follow your steps and create necessary iceberg tables for your review.

PR #555, opened 2025-04-14
Update actions cache

ian-r-rose:
Part of fixing https://github.com/cagov/caldata-dse-internal-tracker/issues/68

Updates `@actions/cache` so that we no longer run the deprecated version, and generally fixes it so it works properly

### DIFF
--- a/.sqlfluff
+++ b/.sqlfluff
@@ -1,0 +1,9 @@
+[sqlfluff]
+# For some reason this can only be set in the root directory, cf
+# https://docs.sqlfluff.com/en/stable/configuration.html#nesting.
+# Other config parameters are set in the dbt project directory, as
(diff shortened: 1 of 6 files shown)


.sqlfluff
```diff
@@ -0,0 +1,10 @@
+[sqlfluff]
+# For some reason this can only be set in the root directory, cf
+# https://docs.sqlfluff.com/en/stable/configuration.html#nesting.
```
JamesSLogan, line 3: I think this link doesn't work. I was curious because it might explain some issues I was having last year...

JamesSLogan (approved):

.sqlfluff
ian-r-rose: Thanks @JamesSLogan! I've updated the link

PR #557, opened 2025-04-29
Use correct target for docs build

### LINKED ISSUE #556: Fix docs build
The docs build is currently failing, seemingly do to a RBAC issue with the CI service account. Should be a quick fix.

ian-r-rose:
Fixes #556, this was an oversight in my recent github actions refresh. We build the docs using the "prd" target so that the docs render with the production schema/database names.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,5 +1,6 @@
 name: docs
 on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


thehanggit (approved):
Thanks @ian-r-rose ! It looks good to me.

Issue #477, opened 2024-11-13
Investigate "microbatch" incremental strategy


ian-r-rose:
dbt 1.9 introduces a new incremental strategy called ["microbatching"](https://docs.getdbt.com/docs/build/incremental-microbatch). It seems like it might be a great fit for our current large incremental time series. It allows some of the following behaviors:

1. Targeted backfills. Previously we had two options: run the most recent data, or do a full table refresh. This allows us to do more specific backfills, like "refresh the last week of data"
2. Less manual configuration of lookbacks. We might be able to remove or significantly simplify `make_model_incremental()`.
4. Breaking large table builds into multiple steps. This probably doesn't matter a huge amount for Snowflake (though I would be interested to be proven wrong!), but with some databases it might help with resource usage.

I think an experimental branch testing this new strategy out would be a great idea. Since it relies on dbt-core 1.9, it would probably need to wait on the version upgrade @summer-mothwood is doing in #352

jkarpen:
Next step is to have a team discussion on this topic.

ian-r-rose:
Okay, I've had a chance to look into this! My experiments (and maybe initial implementation?) are [here](https://github.com/cagov/caldata-mdsa-caltrans-pems/compare/main...microbatch?expand=1).

### Implementing a microbatch model

This is largely done at the configuration level. We need to set an `event_time` config parameter on all microbatch models _as well as_  their upstream models, which may or may not be incremental. This is the column that is used for constructing time-based batches of data. In this project, we would probably use `sample_date` (or `sample_timestamp`) for most event times, and construct daily batches.

We also set a `begin` config parameter that lets the microbatch logic know the start of time when building the model. This is used when initially building the model. It's also an [opportunity](https://docs.getdbt.com/reference/resource-configs/begin#set-begin-to-use-relative-dates) for us to do different things in dev and prod. I'm fairly certain that we could use this to set the dev lookback that `make_model_incremental` currently does.

Finally, there are backfill config parameters `--event-time-start` and `--event-time-end`. These are used for running explicit backfills of microbatch incremental tables. For instance, if there is an outage for a couple of weeks, we could reprocess the data for that period and only that period. Previously we would have had to run a `--full-refresh` of the table for all time to handle that case.

### Dev workflow for microbatch models

Notes on a dev workflow when running microbatch models:

1. When a microbatch model is built for the first time in dev, the default behavior will be to build the last week of data, per the `begin` behavior
[...]

PR #558, opened 2025-05-06
Microbatch implementation

ian-r-rose:
Follow on work from #477 

I've been looking at using the "microbatch" incremental strategy for our large incremental models, and think it would be a nice simplification to a lot of the patterns here (such as being able to delete `make_model_incremental`) as well as enabling easier backfills of missing data. More thoughts on this are in #477. In this PR I've been taking an initial look at implementation, and wanted to try to get some thoughts from @thehanggit before he leaves Caltrans.

For most of our incremental models converting to the microbatch strategy is pretty straightforward (see, e.g., [here](https://github.com/cagov/caldata-mdsa-caltrans-pems/compare/microbatch?expand=1#diff-6ac1dc2e2684615853e30fe5d183f54e8fbc144878308ee09297d6cca5ac4e0a)). But the outlier removal model required a bit more thought, because it always compares with the last seven days' data to identify outliers. I don't really like this approach for a few reasons:

* It's not _idempotent_, in that if I run the model on different days, I'll get different results.
*  It won't work if a station has been decommissioned, or otherwise hasn't produced any data for the last week.
* It doesn't really fit well into the microbatch execution model, where we want to be able to run different days' data independently of each other (this is more of a conceptual cleanliness issue than a technical one).

So, the bulk of this PR is actually doing some refactoring of some of the early data processing for the clearinghouse data. Specifically, I do the following:

1. Compute detector statistics at regular (quarterly) intervals for use in outlier removal. This is similar to what we do for the regression coefficients. When identifying outliers, we use an `asof` join to get the most recent statistics for a detector.
2. Consolidate several of the data cleaning/preparation steps into a single model, including
    * Removing outliers
    * Filling out missing timestamps using the `timestamp_spine` macro that @JamesSLogan created
    * Attaching station metadata (still WIP)
3. Move the computation of the "g factor" speed to after this data preparation step (still WIP). Previously this was done after outlier removal, but before the timestamp spine. Conceptually, it made more sense to me to compute this derived data column after the data prep, but perhaps there is some reason to do it before? @thehanggit I would appreciate any thoughts you have here.


transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
ian-r-rose, line 1: This model is consolidated into `int_vds__detector_agg_five_minutes_normalized`

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
```diff
@@ -1,106 +0,0 @@
...
-        coalesce(agg.direction, dmeta.direction) as direction,
-        coalesce(agg.length, dmeta.length) as length,
-        coalesce(agg.station_valid_from, dmeta._valid_from) as station_valid_from,
-        coalesce(agg.station_valid_to, dmeta._valid_to) as station_valid_to
-    from spine
-    left join detector_agg as agg
-        on spine.timestamp_column = agg.sample_timestamp and spine.detector_id = agg.detector_id
-
-    -- The previously "missing" rows will need metadata filled in
-    left join detector_meta as dmeta
```
ian-r-rose, line 91: I don't think this join is necessary, as the `spine` can include the metadata by itself (which is what I've done in `int_vds__detector_agg_five_minutes_normalized`. Does that seem correct to you @JamesSLogan?

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
```diff
@@ -1,106 +0,0 @@
...
-        on
-            agg.sample_ct is null -- this filters for missing rows since it is a computed value in upstream models
-            and spine.detector_id = dmeta.detector_id
-            and to_date(spine.timestamp_column) >= dmeta._valid_from
-            and (
-                to_date(spine.timestamp_column) < dmeta._valid_to
-                or dmeta._valid_to is null
-            )
-    left join gfactor_speed as gs
```
ian-r-rose, line 100: I believe if we compute the `gfactor_speed` _downstream_ of this model, we don't have to do this join,can you think of any reason to not do that @thehanggit ?

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
```diff
@@ -1,106 +0,0 @@
...
-        agg.zero_vol_ct,
-        agg.occupancy_avg,
-        agg.zero_occ_ct,
-        agg.zero_vol_pos_occ_ct,
-        agg.zero_occ_pos_vol_ct,
-        agg.high_volume_ct,
-        agg.high_occupancy_ct,
-        coalesce(agg.speed_weighted, gs.imputed_speed) as speed_weighted,
-        agg.volume_observed,
-        coalesce(agg.state_postmile, dmeta.state_postmile) as state_postmile,
```
ian-r-rose, line 73: Since these metadata ultimately come from the same source, I don't think we need all of these `coalesce` statements.

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_outlier_agg_five_minutes.sql
```diff
@@ -1,113 +0,0 @@
...
-    select
-        detector_id,
-        avg(volume_sum) as volume_mean,
-        stddev(volume_sum) as volume_stddev,
-        -- consider using max_capacity
-        percentile_cont(0.95) within group (order by volume_sum) as volume_95th,
-        percentile_cont(0.95) within group (order by occupancy_avg) as occupancy_95th
-    from filtered_five_minute_agg_lastweek
-    group by detector_id
-),
```
ian-r-rose, line 64: This (and above) is what I've tried to re-implement in an idempotent way in `int_diagnostic__detector_outlier_thresholds`

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_outlier_agg_five_minutes.sql
```diff
@@ -1,113 +0,0 @@
...
-
--- retrieve recent five-minute data
-five_minute_agg as (
-    select *
-    from {{ ref('int_clearinghouse__detector_agg_five_minutes') }}
-    where {{ make_model_incremental('sample_date') }}
-),
-
--- impute detected outliers
-outlier_removed_data as (
```
ian-r-rose, line 74: This is consolidated into `int_vds__detector_agg_five_minutes_normalized`

transform/models/intermediate/diagnostics/int_diagnostics__constant_occupancy.sql
ian-r-rose, line 1: Nice example of the simple case for microbatch

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
JamesSLogan: Yes, that seems correct! I think I remember originally pulling from a model without metadata and must have not realized...

transform/models/intermediate/vds/int_vds__detector_agg_five_minutes_normalized.sql
JamesSLogan, line 1: If my understanding is correct, this model should have a `five_minute_daily_count` data_test configured

JamesSLogan (commented):
🙌 very nice @ian-r-rose ! I don't see any glaring issues in the models.

transform/models/intermediate/clearinghouse/int_clearinghouse__detector_agg_five_minutes_with_missing_rows.sql
thehanggit: You are right! It is not necessary to join here, as we can join anywhere downstream.

ian-r-rose:
> For the gspeed, I have talked with Ken today and the only reason we brought onto the table is about the performance (Please check PR https://github.com/cagov/caldata-mdsa-caltrans-pems/pull/502). 1. In the 5-minute dataset with missing rows, it would make the table much larger compared to the dataset without missing rows. I'm not sure if it will cause significant delay. 2. The dataset with missing rows generates a large amount of NULL value in flow and occ. I checked the gfactor speed calculation and it should have avoided to include them during calculation. But it still remains unclear how would they affect the gfactor calculation. I would think of a QC to double check the difference. Other than that, there is no other reasons.

My suspicion is that the extra cost of computing gspeed for the missing rows is less than the cost of joining two large tables. In general, joins are more expensive and less parallelizable than operations that can happen on a single row, and filling missing rows seems to mostly increase the table sizes by ~10%.

thehanggit:
> > For the gspeed, I have talked with Ken today and the only reason we brought onto the table is about the performance (Please check PR #502). 1. In the 5-minute dataset with missing rows, it would make the table much larger compared to the dataset without missing rows. I'm not sure if it will cause significant delay. 2. The dataset with missing rows generates a large amount of NULL value in flow and occ. I checked the gfactor speed calculation and it should have avoided to include them during calculation. But it still remains unclear how would they affect the gfactor calculation. I would think of a QC to double check the difference. Other than that, there is no other reasons.
> 
> My suspicion is that the extra cost of computing gspeed for the missing rows is less than the cost of joining two large tables. In general, joins are more expensive and less parallelizable than operations that can happen on a single row, and filling missing rows seems to mostly increase the table sizes by ~10%.

I agree with your claim. Please feel free to move the gspeed model after data prep.

transform/models/intermediate/diagnostics/int_diagnostics__detector_outlier_thresholds.sql
```diff
@@ -0,0 +1,104 @@
...
+    inner join agg_dates_to_evaluate
+        on
+            agg.sample_date >= agg_dates_to_evaluate.agg_date
+            -- TODO: use variable for agg window
+            and agg.sample_date
+            < dateadd(day, {{ var("outlier_agg_time_window") }}, agg_dates_to_evaluate.agg_date)
+    inner join good_detectors
+        on
+            agg.detector_id = good_detectors.detector_id
+            and agg.sample_date = good_detectors.sample_date
```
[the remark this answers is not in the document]
ian-r-rose: Thanks for the suggestion @JamesSLogan! I'm not familiar with this particular style of filter joins, can you give an example?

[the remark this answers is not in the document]
JamesSLogan: I think something like below would perform the filter in a similar manner as the current join to `good_detectors`. The idea is that the database doesn't have to build a large intermediate result set from the join and can instead simply search `good_detectors`, stopping as soon as matches are found. This, of course, depends on how snowflake optimizes the query...

```sql
select
    ...
from agg
inner join agg_dates_to_evaluate
    on
        agg.sample_date >= agg_dates_to_evaluate.agg_date
        and agg.sample_date
        < dateadd(day, {{ var("outlier_agg_time_window") }}, agg_dates_to_evaluate.agg_date)
where
    agg.station_type in ('ML', 'HV')
    and exists (
        select 1
        from good_detectors gd
        where
            gd.detector_id = agg.detector_id
            and gd.sample_date = agg.sample_date
    )
```

There could be a difference in outputted rows if `agg` -> `good_detectors` is one -> many, which I don't think is the case here.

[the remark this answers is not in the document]
ian-r-rose: Interesting...would this be considered a correlated subquery? I've never quite felt smart enough for those...

I'm going to test this to see if I can find a difference in the query plan and performance, thanks!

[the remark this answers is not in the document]
JamesSLogan: Yes, and I feel the same! They're typically harder to read but this case would better convey intent (slightly) and maybe even improve performance 🤞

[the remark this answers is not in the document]
ian-r-rose: I took a look at this, and just documenting what I saw here:

There is almost no difference in performance between these two approaches: the query plans look almost identical, and both involve effectively the same join. This particular join is a small fraction of the total runtime of the query. There are two minor differences I noticed in the query plan:

* First, the join type in the query plan for the correlated subquery is "semi" instead of "inner". From what I understand searching around, a subquery used in this way is effectively synonymous with "semi-join"
* Second, there is a pre-aggregation in the correlated subquery to ensure there is a single `detector_id` and `sample_date` for the existence check. If I understand your comment above, this is what you were referring to above for the one->many join.

Here is a snapshot of the inner join query plan:

image

And here is the correlated subquery:

image

Happy to take this suggestion to convey the intent of the join, thanks for the knowledge sharing @JamesSLogan!

PR #572, opened 2025-07-23
Apply updates from template

ian-r-rose:
Principally migrating to `uv`


ram-kishore-odi (approved):
The changes look good in general. Following are a few items that I was thinking about. Please see below - 
- Do you think pinning dependencies with exact versions in pyproject.toml might be helpful ? 
- Having a note to clarify why Python 3.13 is required might be helpful (it’s very new and might not be universally supported yet).
- In docs/setup.md, clarifying when to use uv run versus activating the virtual environment might be helpful (if there are any key items, I see that you already mentioned about personal preference)
- In drop_schemas_with_prefix.sql, would adding a confirmation step before dropping schemas for added safety help ?

ian-r-rose:
> The changes look good in general. Following are a few items that I was thinking about. Please see below -
> 
>     * Do you think pinning dependencies with exact versions in pyproject.toml might be helpful ?
> 
>     * Having a note to clarify why Python 3.13 is required might be helpful (it’s very new and might not be universally supported yet).

It's not required for any of the functionality: I think it's important that we are consistent about which version we are using, but the specific version isn't that critical, so long as our dependencies support it. We were previously on 3.10, which reaches end-of-life in [about a year](https://devguide.python.org/versions/), and I was just using this as an opportunity to bump the version to the most recent stable release (out since October).

> 
>     * In docs/setup.md, clarifying when to use uv run versus activating the virtual environment might be helpful (if there are any key items, I see that you already mentioned about personal preference)

I think the discussion about the tradeoffs is there, it's right below the note about it being a personal preference. Is there anything more you would want to see there?

> 
>     * In drop_schemas_with_prefix.sql, would adding a confirmation step before dropping schemas for added safety help ?

I'm not sure if dbt macros can do this, I can't see any documentation around waiting for user input.

ram-kishore-odi:
Thank you so much for the responses ! I have no further comments.

PR #584, opened 2025-07-31
Copier updates

MACKAT05:
Applied copier update. and merged elements where needed.

There were some issues with the _macro document patch file. the template names don't match in the template. we will need to revisit that and fix the function names.... its also advisable to name the macro files in line with the macro that they contain.
also i ended up updating several of the python versions .... @ian-r-rose   should we be seeing these updates flow through copier?


transform/models/intermediate/vds/_vds.yml
```diff
@@ -113,27 +113,18 @@ models:
...
   - name: int_vds__detector_config
     description: |
       Metadata for a single loop detector. This is the device that actually records
       flow, occupancy, and speed, and is typically installed in a single lane.
       Multiple detectors across a set of lanes constitute a station.
-
-      Unlike the equivalent configuration models for controller and station,
-      this metadata model is further joined with metadata for each detector's
-      station. This adds some further denormalization to the table, but is a
-      practical response to so many downstream metrics relying on station metadata.
```
ian-r-rose, line 125: Why remove this note? It still seems like useful information

transform/models/intermediate/vds/_vds.yml
```diff
@@ -113,27 +113,18 @@ models:
...
       This table is a slowly-changing-dimension-2, meaning that it has multiple
       different versions of the metadata for each controller, with `_valid_from`
       and `_valid_to` datetime columns indicating when the version is correct.
-
       This model is unique at the level of DETECTOR_ID + _VALID_TO.
-    data_tests:
-      - dbt_utils.unique_combination_of_columns:
-          combination_of_columns:
-            - DETECTOR_ID
-            - _VALID_TO
```
ian-r-rose, line 136: Here as well, this seems like a useful test

ian-r-rose (commented):
This is looking great, I'm excited to have some order restored here. Just a couple of small notes:

1. There is still a merge conflict in the `.sqlfluffignore`
2. I see in a number of places you have commented out column names and descriptions (presumably because those tables used to have those columns, but no longer). From my perspective it's better to remove those all-together rather than leaving dead code commented out.

transform/models/intermediate/vds/_vds.yml
MACKAT05: Likely a victim of copilot will undo.

MACKAT05:
It looks like a number of tests got imputed with the documentation based on tests present in other models... if it is a performance concern we can probably drop them.

MACKAT05:
@JamesSLogan  can you look this over while ian is out.

transform/models/intermediate/clearinghouse/_clearinghouse.yml
```diff
@@ -311,3 +424,89 @@ models:
...
+      - name: volume_sum
+        description: |
+          For the 5-minute aggregation there should be 10 30-second samples collected. If 10 or more
+          samples are collected the aggregated (sum) volume rounded to the nearest integer value is used
+          but if less than 10 samples are recieved in the 5-minute timeframe the flow is normalized using
+          the equation: 10 / sample count * sum(volume).
+      - name: zero_vol_ct
+        description: Counts the number of raw data samples where a lane's volume (flow) value equals 0.
+      - name: occupancy_avg
+        description: Occupancy of the lane during the sample period expressed as a decimal number between
```
JamesSLogan, line 453: sentence was truncated, should be `...between 0 and 1` I believe

transform/models/marts/geo/geo.yml
```diff
@@ -0,0 +1,112 @@
+version: 2
+
+models:
+  - name: geo__current_detectors
+    description: |
+      Contains metadata for all current detectors, including location and configuration details.
+    columns:
+      - name: geometry
+        description: "Spatial point geometry for each detector."
```
JamesSLogan, line 9: A bit of a nit, but could you remove the quotes from here and the rest of the file to keep things consistent?

JamesSLogan (commented):
This appears nearly ready for merge - there is one [comment ](https://github.com/cagov/caldata-mdsa-caltrans-pems/pull/584#discussion_r2248259975)from Ian that needs to be addressed still.

I'm neutral on the new data tests - we'll of course want to keep an eye on them post-merge.

transform/models/intermediate/vds/_vds.yml
MACKAT05: this should be fixed in b018bc320c84a6d13441506bff7c7f9389b3a8e3

MACKAT05:
@JamesSLogan   we got a new and interesting failure since i added the check to ci... do we need to adjust the ci build to build all or just delete the item from the ci build...

JamesSLogan:
I think I see what's going on - this PR only has documentation changes so nothing gets built by `dbt build --select state:modified+ --defer`, and therefore there is nothing for the macro to validate. This PR feels like an edge case but even in more normal cases we'll run into this issue since it'll be exceedingly rare to update _every_ model.

We still want CI to only check changed models to keep CI runs short (they used to be 45 minutes). I think we need to enhance our macro instead, here are a few ideas:

1. Enhance our macro to respect the `--select state:modified+` argument. [dbt documentation ](https://docs.getdbt.com/reference/commands/run-operation)doesn't indicate that `--select` is valid in a macro but my CLI does when I run `dbt run-operation --help`
2. Add an argument to the macro to ignore completely missing tables and enable it in CI
3. Add an argument to the macro that tells it what nodes to parse...this is from an LLM but could work:
    ```bash
    dbt run-operation validate_all_schemas \
    --args '{"models": {{ dbt list --select state:modified+ --output name | join(",") }} }'
    ```

transform/models/marts/imputation/_imputation.yml
```diff
@@ -80,11 +74,6 @@ models:
       This model filters the results of the imputation_detector_imputed_agg_five_minutes model from the transform
       step to detectors with station_type in of either 'ML' or 'HV' from the past 4 days and adds the county name
       to the results. This model is unique at the level of SAMPLE_TIMESTAMP + DETECTOR_ID.
-    data_tests:
-      - dbt_utils.unique_combination_of_columns:
-          combination_of_columns:
-            - SAMPLE_TIMESTAMP
-            - DETECTOR_ID
```
ian-r-rose, line 87: Why remove this test?

ian-r-rose (approved):
Looks good to me, just one question about a removed test

ian-r-rose:
> It looks like a number of tests got imputed with the documentation based on tests present in other models... if it is a performance concern we can probably drop them.

This *could* be a problem in some cases, as many of the tables are quite large. In particular, uniqueness tests can get us into some trouble. Let's monitor

transform/models/marts/imputation/_imputation.yml
ian-r-rose: @MACKAT05

